PR #87, opened 2021-03-29
Forbid create voucher from diff type of supply

HristiyanG:


thcrnk (approved):

hswopeams (approved):
Changes made to error codes will break the front-end after the contracts are redeployed. Discussed with Nassar and Hris on 2 april 2021. Since front-end doesn't use mocks, it will need the changed contract to test against. Approving this PR with the caveat that the front-end needs to be changed ASAP to match the error codes.

test/2_test_fullpath_with_permit.js
```diff
@@ -1793,6 +1812,25 @@ contract('Cashier and VoucherKernel', async (addresses) => {
           );
         });
 
+        it('[NEGATIVE] Should not create order from a wrong payment type', async () => {
```
ali2251, line 1815: Whats the difference between these tests? they all look exactly the same to me unless i am missing something?

contracts/BosonRouter.sol
```diff
@@ -50,21 +50,21 @@ contract BosonRouter is
      * @notice Acts as a modifier, but it's cheaper. Checking if a non-zero address is provided, otherwise reverts.
      */
     function notZeroAddress(address tokenAddress) private pure {
-        require(tokenAddress != address(0), "0_A"); //zero address
+        require(tokenAddress != address(0), "0A"); //zero address
```
ali2251, line 53: Are these changes needed for this PR?

contracts/VoucherKernel.sol
```diff
@@ -310,15 +310,20 @@ contract VoucherKernel is IVoucherKernel, Ownable, Pausable, UsingHelpers {
...
      * @param _issuer          Address of the token's issuer
      * @param _holder          Address of the recipient of the voucher (ERC-721)
+     * @param _paymentMethod   method being used for that particular order that needs to be fulfilled
      * @param _correlationId           ID of the current interaction with the smart contract for a specific user
      */
     function fillOrder(
         uint256 _tokenIdSupply,
         address _issuer,
         address _holder,
+        uint8 _paymentMethod,
```
ali2251, line 320: where is this payment method used?

contracts/BosonRouter.sol
HristiyanG: Yes. I think this was discussed last week (Thursday) on the call we had.

contracts/VoucherKernel.sol
HristiyanG: The payment method is a substantial part throughout the whole business logic.

test/2_test_fullpath_with_permit.js
HristiyanG: Maybe you should have a look at the Builder

PR #96, opened 2021-04-21
Unit test changes 1 test fullpath

hswopeams:
This is the PR for the "Upgrade unit tests in 1_test_fullpath.js" story

### DIFF
--- a/test/1_test_fullpath.js
+++ b/test/1_test_fullpath.js
@@ -1067,7 +1067,8 @@ contract('Voucher tests', async (addresses) => {
         'LogFundsReleased',
         (ev) => {
           return (
-            ev._tokenIdVoucher.eq(tokenVoucherKey1) && ev._type.eq(constants.ZERO)
+            ev._tokenIdVoucher.eq(tokenVoucherKey1) &&
+            ev._type.eq(constants.ZERO)
           );
         },
         'funds not released successfully'
--- a/test/1_test_fullpath.js
+++ b/test/1_test_fullpath.js
@@ -171,24 +171,24 @@ contract('Voucher tests', async (addresses) => {
       //Check VocherKernel State
       const promise = await contractVoucherKernel.promises(promiseId1);
       assert.equal(promise.promiseId, promiseId1, "Promise Id incorrect");
-      promise.nonce.eq(new BN(1));
+      assert.isTrue(promise.nonce.eq(new BN(1)),'Nonce is incorrect');
       assert.strictEqual(promise.seller, users.seller.address, "Seller incorrect");
-      promise.validFrom.eq(new BN(constants.PROMISE_VALID_FROM));
-      promise.validTo.eq(new BN(constants.PROMISE_VALID_TO));
-      promise.price.eq(new BN(constants.PROMISE_PRICE1));
-      promise.depositSe.eq(new BN(constants.PROMISE_DEPOSITSE1));
-      promise.depositBu.eq(new BN(constants.PROMISE_DEPOSITBU1));
-      promise.idx.eq(new BN(1));
+      assert.isTrue(promise.validFrom.eq(new BN(constants.PROMISE_VALID_FROM)));
+      assert.isTrue(promise.validTo.eq(new BN(constants.PROMISE_VALID_TO)));
+      assert.isTrue(promise.price.eq(new BN(constants.PROMISE_PRICE1)));
+      assert.isTrue(promise.depositSe.eq(new BN(constants.PROMISE_DEPOSITSE1)));
+      assert.isTrue(promise.depositBu.eq(new BN(constants.PROMISE_DEPOSITBU1)));
+      assert.isTrue(promise.idx.eq(new BN(0)));
 
       const orderPromiseId = await contractVoucherKernel.ordersPromise(tokenSupplyKey1);
       assert.strictEqual(orderPromiseId, promiseId1, "Order Promise Id incorrect");
 
       const tokenNonce = await contractVoucherKernel.tokenNonces(users.seller.address);
-      tokenNonce.eq(new BN(1));
+      assert.isTrue(tokenNonce.eq(new BN(1)));
 
       //Check ERC1155ERC721 state
       const sellerERC1155ERC721Balance = await contractERC1155ERC721.balanceOf(users.seller.address, tokenSupplyKey1);
-      sellerERC1155ERC721Balance.eq(new BN(1));
+      assert.isTrue(sellerERC1155ERC721Balance.eq(new BN(1)));
  
     });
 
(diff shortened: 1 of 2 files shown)


test/1_test_fullpath.js
```diff
@@ -23,13 +23,16 @@ contract('Voucher tests', async (addresses) => {
...
-  let tokenSupplyKey1, tokenSupplyKey2, tokenVoucherKey1, tokenVoucherKey2;
+  let tokenSupplyKey1,
+    tokenSupplyKey2,
+    tokenVoucherKey1,
+    tokenVoucherKey2,
+    promiseId1,
+    promiseId2;
 
-  before('setup contracts for tests', async () => {
+  beforeEach('setup contracts for tests', async () => {
```
HristiyanG, line 33: Do we need to redeploy the contracts before each test? Can we try to avoid that where possible, so we can speed up the tests a little bit?

test/1_test_fullpath.js
```diff
@@ -84,14 +87,881 @@ contract('Voucher tests', async (addresses) => {
...
+
+      let tokenSupplyKey1;
+
+      truffleAssert.eventEmitted(
+        txOrder,
+        'LogOrderCreated',
+        (ev) => {
+          tokenSupplyKey1 = ev._tokenIdSupply;
+          return (
+            ev._tokenIdSupply.gt(new BN(0)) &&
```
HristiyanG, line 122: since `new BN(0)` && `new BN(1)` are thoroughly used, can we use constants for them?

test/1_test_fullpath.js
```diff
@@ -84,14 +87,881 @@ contract('Voucher tests', async (addresses) => {
...
+          from: users.attacker.address,
+        }),
         truffleAssert.ErrorType.REVERT
       );
     });
   });
 
-  describe('Voucher Sets (ERC1155)', () => {
-    it('adding one new order / promise', async () => {
+  //HS:  All other withdraw functions are tested in 3_withdrawals.js. Do we want to move this one?. Withdrawal of deposit not included here
```
HristiyanG, line 961: IMHO moving them into 3_withdrawals.js would be more appropriate

test/1_test_fullpath.js
```diff
@@ -84,14 +87,881 @@ contract('Voucher tests', async (addresses) => {
...
+  describe('Withdrawals', function () {
+    beforeEach('setup contracts for tests', async () => {
```
HristiyanG, line 963: Do we actually setup contracts here?

test/1_test_fullpath.js
```diff
@@ -84,14 +87,881 @@ contract('Voucher tests', async (addresses) => {
...
+      const sellerERC1155ERC721BalanceVoucherSet2 = await contractERC1155ERC721.balanceOf(
+        users.seller.address,
+        tokenSupplyKey2
+      );
+      assert.isTrue(sellerERC1155ERC721BalanceVoucherSet2.eq(new BN(1)));
+    });
+  });
+
+  describe('Commit to buy a voucher (ERC1155)', () => {
+    beforeEach('setup contracts for tests', async () => {
```
HristiyanG, line 385: Do we setup the contracts here?

test/1_test_fullpath.js
```diff
@@ -84,14 +87,881 @@ contract('Voucher tests', async (addresses) => {
...
+            value: 0,
+          }
+        ),
+        truffleAssert.ErrorType.REVERT
+      );
+    });
+  });
+
+  describe('Vouchers (ERC721)', function () {
+    beforeEach('setup contracts for tests', async () => {
```
HristiyanG, line 720: We might need to change the comment in the `beforeEach`  where possible

test/1_test_fullpath.js
```diff
@@ -166,181 +1023,100 @@ contract('Voucher tests', async (addresses) => {
...
       );
-      const internalTx = await truffleAssert.createTransactionResult(
-        contractVoucherKernel,
-        txFillOrder.tx
+
+      const sellerBalanceBefore = await new BN(
+        await web3.eth.getBalance(users.seller.address)
       );
 
+      console.log('sellerBalanceBefore ', sellerBalanceBefore.toString());
```
HristiyanG, line 1047: Do we need this `console.log` ?

HristiyanG (changes requested):

HristiyanG:
Just a general one: 
Do we need to see addresses for the `Seller`, `Buyer`, and `Attacker` before each test? I am not sure if this is of such importance?
Another thing that happened to me was that when running the pipeline locally this test `refunding one voucher, then complain, then cancel/fault` failed once. It happened only once, run the pipeline several times more and it passes, but i think it worth mentioning it

test/1_test_fullpath.js
hswopeams: We want to be able to run each test case (`it`) using `it.only` in addition to running the test cases as part of the whole script. For this to work, each test case needs to be independent of any other test case. It can't rely on  or be influenced by state changes from previous test cases, so fresh state is needed for each test case. Without fresh state for each test case, we run into problems, especially with test cases that check nonces and correlation Ids. The assertions will pass or fail depending on whether you ran them with `it.only` or as part of the whole script. If the test case has fresh state, the assertions will always pass because the values will always be the same (unless a bug is introduced in the code).

Redeploying each time does take longer, but you shouldn't have to run the full script but once for each change. For example, if you add a new function, you might have a happy path test case and an error/negative test case. You should be able to add your first test case as an `it` and test it using `it.only`. Then you move on to your second test case and do the same. Only after you are satisfied that they both work do you have to run the whole script to run all the unit tests.

test/1_test_fullpath.js
hswopeams: Yes. Good idea. I'll change that.

test/1_test_fullpath.js
hswopeams: OK. Le'ts discuss during the standup if we want to do that now or pick it up when we are refactoring the other file and moving to TypeScript (which I believe is the plan).

test/1_test_fullpath.js
hswopeams: Changed to "perform prerequisite steps"

test/1_test_fullpath.js
hswopeams: Changed to "perform prerequisite steps". Will be included in next commit.

test/1_test_fullpath.js
hswopeams: Changed to "execute prerequisite steps". Will be included in next commit.

test/1_test_fullpath.js
hswopeams: Oops. Nope. Will be removed in next commit.

hswopeams:
> Just a general one:
> Do we need to see addresses for the `Seller`, `Buyer`, and `Attacker` before each test? I am not sure if this is of such importance?
> Another thing that happened to me was that when running the pipeline locally this test `refunding one voucher, then complain, then cancel/fault` failed once. It happened only once, run the pipeline several times more and it passes, but i think it worth mentioning it

Removed the console.log statements that output the  addresses. 

Did the test case fail when you ran it as part of the script or independently with `it.only`?

HristiyanG:
> Removed the console.log statements that output the addresses.
> 
> Did the test case fail when you ran it as part of the script or independently with `it.only`? And do you remember what the error was?

It happened just once. I tried to reproduce running only this test, or run the pipeline couple of times more but it succeeded. Standard error -  did not meet the assertion.

hswopeams:
> > Removed the console.log statements that output the addresses.
> > Did the test case fail when you ran it as part of the script or independently with `it.only`? And do you remember what the error was?
> 
> It happened just once. Failed when run the pipeline. I tried to reproduce running only this test, or run the pipeline couple of times more but it succeeded. Standard error - did not meet the assertion.

I cloned the branch after the latest commit and ran `./go`. It ran fine. It might have been a strange blip. Let's just keep an eye on it. Let me know if you have any problems with it in the future.

test/1_test_fullpath.js
hswopeams: We forgot to talk about it during the standup. I just talked to Ali about it. Let's agree to move it in  the  3_withdrawals.js story. I'll add it as a comment in the story.

HristiyanG (approved):

PR #97, opened 2021-04-21
Change setChange function name to determineStatus and modify calling functionsP

hswopeams:
PR for "Change UsingHelper.setChange to determineStatus" story.

The name setChange implies that a value is being set, which is not the case. The function name is confusing. This function name should be changed to determineStatus.  Calling functions will also need to be changed.


HristiyanG (approved):

ali2251 (approved):
LGTM

hswopeams:
Merge after approval

PR #98, opened 2021-04-28
Move hooks out of router

hswopeams:
PR for story "Move hooks out of BosonRouter into Cashier contract"

Changes to contracts, migration scripts (new references from contracts to other contracts and removal of references required changes to deployment scripts), and tests. Also added a new test file to test new functions such as `Cashier.setBosonRouterAddress`,which didn't have a clear place in any of the current test files. Also moved existing test cases that already set address references on contracts and removed duplicates.

_beforeERC1155Transfer was removed because it included several calls that in the end just access the `balances` mapping in the calling contract(ERC1155721.sol).

### DIFF
--- a/contracts/Cashier.sol
+++ b/contracts/Cashier.sol
@@ -895,8 +895,8 @@ contract Cashier is ICashier, UsingHelpers, ReentrancyGuard, Ownable, Pausable {
             );
 
             //Reduce _from escrow token amount and increase _to escrow token amount - deposit
-            escrowTokens[tokenAddress][_from] =  escrowTokens[tokenAddress][_from].sub(depositBu);
-            escrowTokens[tokenAddress][_to] =  escrowTokens[tokenAddress][_to].add(depositBu);
+            escrowTokens[tokenAddress][_from] = escrowTokens[tokenAddress][_from].sub(depositBu);
+            escrowTokens[tokenAddress][_to] = escrowTokens[tokenAddress][_to].add(depositBu);
 
         }
 
@@ -907,8 +907,8 @@ contract Cashier is ICashier, UsingHelpers, ReentrancyGuard, Ownable, Pausable {
         
 
             //Reduce _from escrow token amount and increase _to escrow token amount - price 
-            escrowTokens[tokenAddress][_from] =  escrowTokens[tokenAddress][_from].sub(price);
-            escrowTokens[tokenAddress][_to] =  escrowTokens[tokenAddress][_to].add(price);
+            escrowTokens[tokenAddress][_from] = escrowTokens[tokenAddress][_from].sub(price);
+            escrowTokens[tokenAddress][_to] = escrowTokens[tokenAddress][_to].add(price);
 
             //Reduce _from escrow amount and increase _to escrow amount - deposit
             escrow[_from] = escrow[_from].sub(depositBu);
@@ -922,16 +922,16 @@ contract Cashier is ICashier, UsingHelpers, ReentrancyGuard, Ownable, Pausable {
 
 
             //Reduce _from escrow token amount and increase _to escrow token amount - price 
-            escrowTokens[tokenAddress][_from] =  escrowTokens[tokenAddress][_from].sub(price);
-            escrowTokens[tokenAddress][_to] =  escrowTokens[tokenAddress][_to].add(price);
+            escrowTokens[tokenAddress][_from] = escrowTokens[tokenAddress][_from].sub(price);
+            escrowTokens[tokenAddress][_to] = escrowTokens[tokenAddress][_to].add(price);
 
             tokenAddress = IVoucherKernel(voucherKernel).getVoucherDepositToken(
                 tokenSupplyId
             );
 
             //Reduce _from escrow token amount and increase _to escrow token amount - deposit 
-            escrowTokens[tokenAddress][_from] =  escrowTokens[tokenAddress][_from].sub(depositBu);
-            escrowTokens[tokenAddress][_to] =  escrowTokens[tokenAddress][_to].add(depositBu);
+            escrowTokens[tokenAddress][_from] = escrowTokens[tokenAddress][_from].sub(depositBu);
+            escrowTokens[tokenAddress][_to] = escrowTokens[tokenAddress][_to].add(depositBu);
 
         }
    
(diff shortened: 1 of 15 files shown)


test/1_test_fullpath.js
```diff
@@ -57,19 +53,31 @@ contract('Voucher tests', async (addresses) => {
     await contractERC1155ERC721.setVoucherKernelAddress(
       contractVoucherKernel.address
     );
-    await contractERC1155ERC721.setBosonRouterAddress(
-      contractBosonRouter.address
-    );
+
+    await contractERC1155ERC721.setCashierAddress(contractCashier.address);
+
+    await contractERC1155ERC721.setCashierAddress(contractCashier.address);
```
ali2251, line 59: is it duplicated?

test/1_test_fullpath.js
```diff
@@ -1148,19 +1151,32 @@ contract('Voucher tests - UNHAPPY PATH', async (addresses) => {
...
```
ali2251, line 1157: same here, seems like its doing the same thing

test/7_ admin_functionality.js
```diff
@@ -0,0 +1,304 @@
...
+
+    it('[NEGATIVE][setTokenContractAddress] Should revert if ZERO address is provided', async () => {
+      await truffleAssert.reverts(
+        contractCashier.setTokenContractAddress(constants.ZERO_ADDRESS),
+        truffleAssert.ErrorType.REVERT
+      );
+    });
+  });
+
+  describe('ERC1125721', () => {
```
ali2251, line 129: 1155 perhaps?

ali2251 (changes requested):

contracts/Cashier.sol
```diff
@@ -826,28 +841,158 @@ contract Cashier is ICashier, UsingHelpers, ReentrancyGuard, Ownable, Pausable {
...
+     * @param _to next owner of the _tokenIdVoucher
+     * @param _tokenIdVoucher _tokenIdVoucher that has been transferred
      */
-    function addEscrowTokensAmount(
-        address _token,
-        address _account,
-        uint256 _newAmount
-    ) external override onlyFromRouter {
-        escrowTokens[_token][_account] =  escrowTokens[_token][_account].add(_newAmount);
+    function onERC721Transfer(
```
HristiyanG, line 858: My understanding was that we move the hooks in a separate contract, responsible only for those validations? Why we need to extend the Cashier?

HristiyanG, line 858: My understanding was that we move the hooks in a separate contract, responsible only for those validations? Why we need to extend the Cashier?

contracts/Cashier.sol
```diff
@@ -826,28 +841,158 @@ contract Cashier is ICashier, UsingHelpers, ReentrancyGuard, Ownable, Pausable {
...
     }
 
+    /**
+     * @notice After the transfer happens the _tokenSupplyId should be updated in the promise. Escrow funds for the seller's deposits (If in ETH) should be allocated to the new owner as well.
+     * @param _from prev owner of the _tokenSupplyId
+     * @param _to nex owner of the _tokenSupplyId
+     * @param _tokenSupplyId _tokenSupplyId for transfer
+     * @param _value qty which has been transferred
+     */
+    function onERC1155Transfer(
```
HristiyanG, line 948: Here as well

contracts/Cashier.sol
```diff
@@ -826,28 +841,158 @@ contract Cashier is ICashier, UsingHelpers, ReentrancyGuard, Ownable, Pausable {
...
+                    _tokenSupplyId
+                );
+
+            depositSe = IVoucherKernel(voucherKernel).getSellerDeposit(
+                _tokenSupplyId
+            );
+            totalAmount = depositSe.mul(_value);
+
+            //Reduce _from escrow token amount and increase _to escrow token amount - deposit
+            escrowTokens[tokenDepositAddress][_from] =  escrowTokens[tokenDepositAddress][_from].sub(totalAmount);
```
HristiyanG, line 985: Two whitespaces after `=`
Please run the pipeline locally, so we avoid that.

contracts/Cashier.sol
```diff
@@ -826,28 +841,158 @@ contract Cashier is ICashier, UsingHelpers, ReentrancyGuard, Ownable, Pausable {
...
+            escrowTokens[tokenDepositAddress][_to] =  escrowTokens[tokenDepositAddress][_to].add(totalAmount);
```
HristiyanG, line 986: Two whitespaces after `=`
Please run the pipeline locally, so we avoid that.

HristiyanG (changes requested):

contracts/Cashier.sol
hswopeams: The hooks were making a lot of calls to the Cashier. Some calls are being made to VouherKernel as well, but most of the external calls were made to Cashier. By moving the hooks to Cashier, the functions can directly acces `escrowTokens` and `escrow` mappings. It removes several external calls per hook function, which makes the hook functions more efficient

HristiyanG: Somewhat agree, but don't feel this is the correct place. You are correct we are making calls to Cashier and VK, but for me we should have dedicated contract responsible for the hooks. @ali2251 Do you have an opinion here?

contracts/Cashier.sol
hswopeams: I did run the pipeline locally -- multiple times. I ran the `./go` script and also `./go contracts:lint`,` ./go contracts:lint_fix`, `./go contracts:format`, `./go contracts:format_fix`, `./go tests:lint`, `./go tests:lint_fix`,` ./go tests:format,` and `./go tests:format_fix` separately.

As you can see, it also passed when the pipeline was run from GitHub. This was never flagged. I will change it manually, but the scripts don't flag everything so some things can slip through.

contracts/Cashier.sol
hswopeams: The hooks were making a lot of calls to the Cashier. Some calls are being made to VouherKernel as well, but most of the external calls were made to Cashier. By moving the hooks to Cashier, the functions can directly acces escrowTokens and escrow mappings. It removes several external calls per hook function, which makes the hook functions more efficient

contracts/Cashier.sol
hswopeams: I did run the pipeline locally -- multiple times. I ran the ./go script and also ./go contracts:lint, ./go contracts:lint_fix, ./go contracts:format, ./go contracts:format_fix, ./go tests:lint, ./go tests:lint_fix, ./go tests:format, and ./go tests:format_fix separately.

As you can see, it also passed when the pipeline was run from GitHub. This was never flagged. I will change it manually, but the scripts don't flag everything so some things can slip through.

contracts/Cashier.sol
ali2251: I think if we reach a point where we are out of space in Cashier, we should create a new contract and move these there but for now it seems the easiest to do it this way

HristiyanG (approved):

ali2251 (approved):

PR #107, opened 2021-05-21
Gas reporter

hswopeams:
This is the PR for the "Gas cost optimisations metrics" card. The eth-gas-reporter was configured. It produces a .txt file when unit tests are run.

The output file isn't very nicely formatted. Outputting to standard out is formatted more nicely, but then you have to make screen shots.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,4 +28,4 @@ coverage*
 
 # Personal Preferences
 .vscode
-gas-reporter.txt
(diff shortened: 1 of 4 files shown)


sriharikapu (commented):
- [x] Works as expected
- [x] Stores the logs in txt file

levalleux-ludo (approved):

PR #106, opened 2021-05-21
Make state variables private

hswopeams:
This is the PR for the "Make state variables private so that default getters functions are not generated" card.
Make state variables private. Needed to add a few functions to facilitate testing. Modified unit tests where they were using the generated getters. Some changes to reference app code will be required, so merging must be coordinated

PR #111, opened 2021-06-09
Add getters for private variables

HristiyanG:
Further to #106 this PR makes state variables private throughout the entire codebase, and expose explicit getters for the one needed in some specific cases as "disaster state", as long as the contract addresses for transparency. @hswopeams, if the document you wrote about that could be applied here, I am happy to include it as additional ref.

### DIFF
--- a/test/3_withdrawals.js
+++ b/test/3_withdrawals.js
@@ -5253,11 +5253,8 @@ contract('Cashier withdrawals ', async (addresses) => {
         truffleAssert.eventEmitted(tx, 'LogDisasterStateSet', (ev) => {
           return ev._triggeredBy == users.deployer.address;
         });
-      });
 
-      it('Disaster State should be set successfully', async () => {
         const disasterState = await contractCashier.isDisasterStateSet();
-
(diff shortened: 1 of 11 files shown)


test/3_withdrawals.js
```diff
@@ -5238,14 +5238,29 @@ contract('Cashier withdrawals ', async (addresses) => {
...
+        );
+        const tx = await cashier.setDisasterState();
 
         truffleAssert.eventEmitted(tx, 'LogDisasterStateSet', (ev) => {
           return ev._triggeredBy == users.deployer.address;
         });
       });
 
+      it('Disaster State should be set successfully', async () => {
+        const disasterState = await contractCashier.isDisasterStateSet();
```
hswopeams, line 5259: Each test case should be independently runnable  (i.g. with `.only`) and should not rely on state set in a previous test case. Change the test case to make it independent.

hswopeams (changes requested):
Please make the specified change to the test.

hswopeams (approved):

PR #113, opened 2021-06-16
Added run task for contracts & updated README & added mock token deployment

dennisfurrer:
This is related to: https://app.asana.com/0/1199560399769920/1200478117062022
AND: https://app.asana.com/0/1199560399769920/1200484274111156

This adds:
 - a package.json script for running the contracts via truffle migrate (using the development network)
 - a rakefile/go task for running the contracts (executes the above package.json script)
 - updates to the README to reflect these changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,15 +15,20 @@ the latest deployment details here, as well.
 For more details about how Boson Protocol works and how you might make use of
 it, please see the [documentation site](https://docs.bosonprotocol.io/).  
 
+---
 **Table of Contents**
 
 - [Local Development](#local-development)
-- [Testing](#testing)
-- [Code Linting](#code-linting)
-- [Documentation](#documentation)
+  - [Prerequisites](#prerequisites)
+  - [Build](#build)
+  - [Run](#run)
+  - [Test](#test)
+  - [Code Linting & Formatting](#code-linting--formatting)
+- [Documentation](#documentation)  
 - [Contributing](#contributing)
 - [License](#license)
 
+---
 ## Local Development
 
 ### Prerequisites
@@ -43,7 +48,8 @@ For instructions on how to get set up with these specific versions:
 * See the [OS X guide](docs/setup/osx.md) if you are on a Mac.
 * See the [Linux guide](docs/setup/linux.md) if you use a Linux distribution.
 
-### Running the build
+---
+### Build
 
 We have a fully automated local build process to check that your changes are
 good to be merged. To run the build:
@@ -68,9 +74,17 @@ To compile the contracts:
 ./go contracts:compile
 ```
 
-## Testing
+---
+### Run
+To deploy a local instance of the contracts run the following command:
+```shell
+./go contracts:run
+```
 
-### Unit Tests
+---
+### Test
+
+#### Unit Tests
 
 All contracts are thoroughly unit tested using 
 [Truffle's JavaScript testing](https://www.trufflesuite.com/docs/truffle/testing/writing-tests-in-javascript) 
@@ -94,7 +108,7 @@ otherwise, create a JSON file creating accounts in the same format as
 ./go "tests:unit[<port>,<path-to-accounts-json>]"
 ```
 
-### Coverage
+#### Coverage
 
 We use [solidity-coverage](https://github.com/sc-forks/solidity-coverage) to 
 provide test coverage reports. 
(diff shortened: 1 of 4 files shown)


hswopeams:
Made a minor change to package.json. Added migration script to deploy mock token locally only. Updated README

README.md
```diff
@@ -68,9 +76,35 @@ To compile the contracts:
...
 
-## Testing
+---
+### Run
+To deploy instances of the contracts for local development without prior knowledge Ganache and Truffle, run the following command:
+```shell
+./go contracts:run
+```
+
+This command starts up Ganache on a random port and migrates all contracts to the Ganache instance. The Ganache instance will remain running the the background, even though the command prompt is available.
```
dennisfurrer, line 86: "running the the background" should be "running in the background"

README.md
```diff
@@ -68,9 +76,35 @@ To compile the contracts:
...
+The pid of the Ganache instance is written to a file in the run/pid directory.
+
+To stop the Ganache instance, run the folloing command:
```
dennisfurrer, line 89: "folloing" should be "following"

README.md
levalleux-ludo, line 86: "even though the command prompt is available" should be "even though the command prompt is not available"?

README.md
```diff
@@ -68,9 +76,35 @@ To compile the contracts:
...
+```shell
+./go ganache:stop
+```
+
+If preferred by those who are familiar with Truffle and Ganache, Ganache can be started up manually in a terminal using the command (or other desired port):
```
levalleux-ludo, line 94: I've understood "truffle migrate" will only work if the current port is 8545, no?
So in case the user needs to choose another port, he/she would have to change the truffle-config.js file accordingly?

package.json
```diff
@@ -7,6 +7,7 @@
     "test": "test"
   },
   "scripts": {
+    "contracts:run": "truffle migrate --network test",
```
levalleux-ludo, line 10: I'm surprised the network used to "run the contract" (means "deploy locally") is not the "development" network but the "test" network.
Meaning that the user has to define PORT/HOST environment variables.
I can't find any mention of that env var in the README.
Did I miss something?

levalleux-ludo (changes requested):

README.md
hswopeams: No. Normally when you startup ganache (unless you specify it should run in the background and/or nohup mode),  there is no command prompt because the process is running in the foreground. When you execute the` ./go contracts:run` command, it starts up Ganache in the background and the command prompt returns. You can't see the process running. There's no indication it is running unless you run the `ps - ef | grep ganache` command

README.md
hswopeams: Yes, but the command to start Ganache just above specifies port 8545. If someone wants to start Ganache on a different port, they can run the migrate command using the test network or specify their own network in truffle.congi.js Anyone who is familiar with Truffle and Ganache will know this. The README isn't meant to be a course in Truffle and Ganache. That's why the `./go contracts:run` is provided for those who don't know how to use it.

package.json
hswopeams: Yes, you missed something. The user doesn't do not have to define any variables. This  variables are set by the `./go` script. Using the test network is consistent with how the `./go tests:unit` task works.

README.md
hswopeams: Will change.

README.md
hswopeams: Will Change

README.md
levalleux-ludo: OK.
I've missunderstood the sentence, thinking that the mentionned command prompt was the command prompt of Ganache-Cli...
Don't you think rephrasing 'event through the command prompt returns" instead of "is available" would help preventing the confusion?

README.md
levalleux-ludo: My comments was referring to the mention "(or other desired port)" at the end of the sentence, that should confuse people because if they decide to run on another port, the next command "truffle migrate" will not work anymore.
So either you can remove this additional mention, or explain what to be changed if the user wants to set another port for the "truffle migrate" to work

package.json
levalleux-ludo: OK

README.md
hswopeams: Changed to "returns".

README.md
hswopeams: Removed (or other desired port)

levalleux-ludo (approved):

hswopeams (approved):

PR #114, opened 2021-07-02
Remove extra using SafeMath line

hswopeams:
The `using SafeMath for uint256;` was duplicated on line 31, probably due to a merge. This PR removes the duplicate line.


levalleux-ludo (changes requested):
You have the same line duplication in MockBosonRouter contract as well
Please change it too

levalleux-ludo (approved):

PR #118, opened 2021-07-08
Local run on windows

levalleux-ludo:
Asana cards: [https://app.asana.com/0/1199560399769920/1200544206392313](https://app.asana.com/0/1199560399769920/1200544206392313)
[https://app.asana.com/0/1199560399769920/1200446754208594](https://app.asana.com/0/1199560399769920/1200446754208594)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,10 +21,14 @@ it, please see the [documentation site](https://docs.bosonprotocol.io/).
 - [Local Development](#local-development)
   - [Prerequisites](#prerequisites)
   - [Build](#build)
-  - [Run](#run)
+  - [Run (Linux / Mac OS X)](#run-linux--mac-os-x)
+  - [Run (Windows)](#run-windows)
   - [Test](#test)
+    - [Unit Tests](#unit-tests)
+    - [Coverage](#coverage)
+    - [Interaction Tests](#interaction-tests)
   - [Code Linting & Formatting](#code-linting--formatting)
-- [Documentation](#documentation)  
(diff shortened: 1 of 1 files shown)


dennisfurrer (approved):

hswopeams (approved):

README.md
```diff
@@ -101,6 +105,20 @@ In a separate terminal, contracts can be deployed using
 ```
 One of the contracts that gets deployed locally is a mock contract that represents the $BOSON token. The mock exists for unit testing purposes and so that those who want to develop against the protocol locally don't have to point to a testnet deployment of the $BOSON token.
 
+---
+### Run (Windows)
+On Windows, the 'go' script is not supported. The standard Truffle and Ganache commands need to be used.
+
+Ganache can be started up manually in a terminal using the command:
+```shell
```
damir-bajramovic, line 113: I don't think that here it should not say `shell`, as `shell` is not being used, but the `cmd` or `Powershell`, as it uses the Windows paths. It is misleading, as, for instance, `git-bash` can be used on Windows environment as well, which uses the UNIX path spec.

levalleux-ludo: 'shell' in the README file is a markdown formatting instruction and does not appear in the rendered file:

HristiyanG (approved):

mischat:
This needs to be revisted since we have moved away from ganache and truffle ... Seems like these docs might now be outdated ? @levalleux-ludo ?

PR #119, opened 2021-07-30
Migrate to ethers

HristiyanG:


test/1_test_fullpath.js
```diff
@@ -124,23 +152,30 @@ contract('Voucher tests', (addresses) => {
 
   describe('Direct minting', function () {
     it('must fail: unauthorized minting ERC-1155', async () => {
-      await truffleAssert.reverts(
-        contractERC1155ERC721.mint(users.attacker.address, 666, 1, []),
-        truffleAssert.ErrorType.REVERT
-      );
+      await expect(
+        contractERC1155ERC721.functions[fnSignatures.mint1155](
```
hswopeams, line 156: Why is the call being done using the function signature instead of just calling the function. This Waffle matcher can be called in the following way:

```
await expect(token.transfer(walletTo.address, 1007))
  .to.be.revertedWith('Insufficient funds');
```

The function can be called directly.

test/1_test_fullpath.js
```diff
@@ -292,16 +313,21 @@ contract('Voucher tests', (addresses) => {
       );
 
       //Check ERC1155ERC721 state
-      const sellerERC1155ERC721Balance = await contractERC1155ERC721.balanceOf(
-        users.seller.address,
-        tokenSupplyKey1
-      );
+      const sellerERC1155ERC721Balance = (
+        await contractERC1155ERC721.functions[fnSignatures.balanceOf1155](
```
hswopeams, line 317: Same question as above. Why is the function signature being used instead of calling the function directly?

package.json
```diff
@@ -40,18 +40,14 @@
...
     "axios": "^0.21.1",
     "chai": "^4.3.4",
     "dotenv": "^8.2.0",
     "elliptic": "^6.5.4",
     "eslint": "^7.18.0",
     "eslint-config-prettier": "^7.2.0",
-    "eslint-plugin-truffle": "^0.3.1",
     "ethereum-waffle": "^3.3.0",
     "ethereumjs-util": "^7.0.7",
     "ethers": "^5.0.32",
```
hswopeams, line 53: It looks like 5.4 is the latest version. Is 5.0.32 the version required by hardhat plug-ins, or can we use a later version?

testHelpers/events.js
```diff
@@ -0,0 +1,89 @@
...
+      } catch (e) {
+        if (e.message.includes('no matching event')) continue;
+        throw new Error(e);
+      }
+    }
+  }
+
+  if (!found) {
+    throw new Error(`Event with name ${eventName} was not emitted!`);
+  }
```
hswopeams, line 81: Isn't using something like 

```
const blockNumber = await ethers.provider.getBlockNumber();
const events = await instance.queryFilter("LogTokenOwnerAddressChanged", blockNumber - 1, blockNumber);
```

a more straight-forward way to get the event? The `instance` in this case is the instance the original function was called on. When using this manner of getting events, I've found I have to check the current block and the one before. Unlike with truffle tests (which usually process transactions from a test case in one block), with hardhat it seems there can be multiple block.

testHelpers/utils.js
```diff
@@ -50,22 +60,32 @@ class Utils {
...
-        from: seller.address,
         value: txValue,
       }
     );
 
-    return returnTx ? txOrder : txOrder.logs[0].args._tokenIdSupply.toString();
+    const txReceipt = await txOrder.wait();
+    let eventArgs;
+
+    events.assertEventEmitted(
```
hswopeams, line 70: Good idea to add this assertion here.

hswopeams (commented):
See comments inline.

package.json
HristiyanG: Yes, we could

test/1_test_fullpath.js
HristiyanG: `transfer` might be called directly, but identical functions with different signatures could not, hence we need to get their signature explicitly

testHelpers/events.js
HristiyanG: By default Hardhat Network is configured to auto-mine blocks, immediately upon receiving each transaction unless there is no specific configuration to change that behaviour - [reference](https://hardhat.org/hardhat-network/explanation/mining-modes.html). I find this way more robust as it does not rely on a block, but on the transaction receipt itself. This way we would not need to know which block this has been included at.  Other way around we might have to guess / increase range on the query filter in case of change of configs

testHelpers/utils.js
HristiyanG: I am not sure if i understood correctly. Assuming you'd like to implement some assertions for the event that was emitted on the particular function - we had a lot of tests covering this scenario in the actual unit testing. This is only a utility file and this function is being used only to  get the arguments from the emitted event. If this is somehow misleading i can implement a new function instead.

hswopeams: There was no event checking in the preview version of the utility.js file. I was just commenting that it was good that it was added.

hswopeams (approved):

PR #123, opened 2021-08-19
Feature/gha

moh-abk:


mischat (changes requested):
I wonder why we choice to upgrade hardhat and as a result a bunch of dependencies, did we need to make those changes to enable the GH actions ?

Furthermore, i think the actions are missing some environment variables. I will pulled in the commits from the main branch and it seems to have failed since.

mischat (dismissed):

dohaki (dismissed):
LGTM

dohaki (approved):
LGTM 👍

mischat (approved):

PR #128, opened 2021-09-07
Remove correlation

hswopeams:
Correlation Id has been removed from all contracts and unit tests.


dennisfurrer (approved):
Looks good to me. Also has @HristiyanG's approval

PR #132, opened 2021-09-20
Multiple token support

hswopeams:
Added ITokenWrapper, DAITokenWrapper, and changed Boson Router to call it


hswopeams:
@levalleux-ludo, @zajck, @mischat Please have a look

contracts/BosonRouter.sol
```diff
@@ -8,10 +8,12 @@ import "@openzeppelin/contracts/utils/Pausable.sol";
 import "@openzeppelin/contracts/utils/Address.sol";
 import "./interfaces/IVoucherKernel.sol";
 import "./interfaces/IERC20WithPermit.sol";
-import "./interfaces/IFundLimitsOracle.sol";
+import "./interfaces/ITokenRegistry.sol";
 import "./interfaces/IBosonRouter.sol";
 import "./interfaces/ICashier.sol";
+import "./interfaces/ITokenWrapper.sol";
 import "./UsingHelpers.sol";
+import "hardhat/console.sol";
```
zajck, line 16: `console` could be removed

contracts/DAITokenWrapper.sol
```diff
@@ -0,0 +1,153 @@
...
+import "hardhat/console.sol";
+
+/**
+ * @title DAITokenWrapper
+ * @notice Contract for wrapping call to DAI token permit function because the DAI token permit function has a different signature from other tokens with which the protocol integrates
+ */
+contract DAITokenWrapper is 
+    ITokenWrapper,
+    Pausable,
+    ReentrancyGuard,
```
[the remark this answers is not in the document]
hswopeams: Removed. Will be included in next commit.

contracts/DAITokenWrapper.sol
```diff
@@ -0,0 +1,153 @@
...
+        bytes32 s
+    ) 
+        external
+        override
+        notZeroAddress(owner)
+        notZeroAddress(spender)
+        whenNotPaused
+    {
+        require(deadline == 0 || block.timestamp <= deadline, "PERMIT_EXPIRED");
+        require(v >= 0 && r != bytes32(0) && s != bytes32(0), "INVALID_SIGNATURE_COMPONENTS");
```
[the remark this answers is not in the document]
hswopeams: Removed. Will be included in next commit.

testHelpers/events.ts
```diff
@@ -30,6 +30,11 @@ export const eventNames = {
   LOG_VK_SET: 'LogVoucherKernelSet',
   LOG_CASHIER_SET: 'LogCashierSet',
   LOG_CANCEL_VOUCHER_SET: 'LogVoucherSetFaultCancel',
+  LOG_TOKEN_WRAPPER_CHANGED: 'LogTokenWrapperChanged',
+  LOG_TOKEN_ADDRESS_CHANGED: 'LogTokenAddressChanged',
+  LOG_PERMT_CALLED_ON_TOKEN: 'LogPermitCalledOnToken',
```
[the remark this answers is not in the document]
hswopeams: Changed.Will be included in next commit.

contracts/BosonRouter.sol
hswopeams: Removed. Will be included in next commit.

zajck:
I reviewed the unit tests and I have the following comments:
* Is testing with waffle mock sufficient? Or should we use the ganache to fork real chain to see the actual behaviour of dai contract? 
* I see that on some parts you have commented out the code following `//Throws data out-of-bounds error....`. I don't know if you actually tried to find the root of the problem. Just in case, when someone will be fixing that: problem occurs, since `IERC721` has 3rd argument `indexed`, while `IERC20WithPermit` does not. So in the former case `data` field of log is empty, while `IERC20WithPermit` expects it, and therfore fails.

Otherwise, tests look ok.

hswopeams:
> I reviewed the unit tests and I have the following comments:
> 
> * Is testing with waffle mock sufficient? Or should we use the ganache to fork real chain to see the actual behaviour of dai contract?
> * I see that on some parts you have commented out the code following `//Throws data out-of-bounds error....`. I don't know if you actually tried to find the root of the problem. Just in case, when someone will be fixing that: problem occurs, since `IERC721` has 3rd argument `indexed`, while `IERC20WithPermit` does not. So in the former case `data` field of log is empty, while `IERC20WithPermit` expects it, and therfore fails.
> 
> Otherwise, tests look ok.

In order to run the unit tests against a forked ganache node (it would probably be better to use a forked hardhat instance since we use hardhat), we would have to completely change the way our pipeline/go script works. I don't think that's feasible in the short timeframe, but is something to consider for the future. Using mocks for unit tests is pretty standard for software development. I'm in favor of using mocks for unit test but having our own test environment (forked ganache or hardhat) to test against before we deploy to Rinkeby.  Unfortunately, we don't have that setup yet. I will see if I can run a separate script/test to run against a forked node, but I don't want to break the pipeline by completely changing how we run unit tests at this point.

Regarding the events, ethers doesn't seem to be able to handle overloaded events. Ethers has a workaround the use of function signatures to call overloaded contract functions. Unfortunately, I couldn't find any way to do something similar with events. I couldn't find any solution online. I kept the code in but commented out so that if the problem is solved in a newer version of ethers, we can uncomment it. Keeping it in but commented also shows we're aware these events should be emitted but can't catch them at the moment for technical reasons.

mischat:
> Is testing with waffle mock sufficient? Or should we use the ganache to fork real chain to see the actual behaviour of dai contract?

This is a great call. In hardhat you can give it a blocknumber and ask it to pull in state of mainnet at block T. And then we can do some testing against the real token. I guess we should create a ticket in the backlog to discuss this. I am also a big fan of tenderley and what that gives you, tenderley + state of mainnet for testing will give us a lot in terms of capabilities.

mischat:
> > > Is testing with waffle mock sufficient? Or should we use the ganache to fork real chain to see the actual behaviour of dai contract?
> > 
> > 
> > This is a great call. In hardhat you can give it a blocknumber and ask it to pull in state of mainnet at block T. And then we can do some testing against the real token. I guess we should create a ticket in the backlog to discuss this. I am also a big fan of tenderley and what that gives you, tenderley + state of mainnet for testing will give us a lot in terms of capabilities.
> 
> Do you agree we should not completely change our pipeline and how we do unit testing right now? We would have to change the pipeline, and it could break. I think this should be a future enhancement, and I think it should be in a test environment that we deploy to before Rinkeby. Rinkeby should be used as the acceptance environment. Mocking in unit tests has been standard in software development for years.

I 100% agree, that not now. 

For now we need to do some sort of manual tests on a deployed version of the DAI contract. I wonder if DAI is deployed to Rinkby?

contracts/DAITokenWrapper.sol
```diff
@@ -0,0 +1,150 @@
...
+     * @param value Number of tokens to be transferred
+     * @param deadline Time after which this permission to transfer is no longer valid
+     * @param v Part of the owner's signatue
+     * @param r Part of the owner's signatue
+     * @param s Part of the owner's signatue
+     */
+    function permit(
+        address owner,
+        address spender,
+        uint256 value,
```
mischat, line 53: This variable is unused. Is the "Number of tokens to be transfered" not needed in the body of this method ?

levalleux-ludo: +1 with the reason given by Heather.
However I can't encourage the idea to have to deal with specificity of each token in the BosonRouter contract (or anywhere in the core protocol). Instead we are trying to have the protocol supporting an unbounded list of tokens (all ERC20 compliants), each specificity being dealt in the Wrapper contract of each token

contracts/DAITokenWrapper.sol
```diff
@@ -0,0 +1,150 @@
...
+        uint nonce =  IDAI(daiTokenAddress).nonces(owner);
+        IDAI(daiTokenAddress).permit(owner, spender, nonce, deadline, true, v, r, s);
+        emit LogPermitCalledOnToken(daiTokenAddress, owner, spender, 0);    
+    }
+
+    /**
+     * @notice Set the address of the wrapper contract for the token. The wrapper is used to, for instance, allow the Boson Protocol functions that use permit functionality to work in a uniform way.
+     * @param _tokenAddress Address of the token which will be updated.
+     */
+    function setTokenAddress(address _tokenAddress)
```
levalleux-ludo, line 76: Any reason why the DAI contract address is not given as a constructor argument?
I guess there is no chance that the DAI contract is changing its address ...

contracts/DAITokenWrapper.sol
```diff
@@ -0,0 +1,150 @@
...
+        external 
+        override
+        onlyOwner
+    {
+        
+        _pause();
+    }
+
+    /**
+     * @notice Unpause the Cashier && the Voucher Kernel contracts.
```
levalleux-ludo, line 115: Is the comment correct?
I'm surprised seeing mentions of the Cashier and VoucherKernel contracts
(same in pause() method comments, above)

contracts/DAITokenWrapper.sol
```diff
@@ -0,0 +1,150 @@
...
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/utils/Pausable.sol";
+
+/**
+ * @title DAITokenWrapper
+ * @notice Contract for wrapping call to DAI token permit function because the DAI token permit function has a different signature from other tokens with which the protocol integrates
+ */
+contract DAITokenWrapper is 
+    ITokenWrapper,
+    Pausable,
```
levalleux-ludo, line 15: Why is that contract Pausable?
The pause does not have any effect on it, and it doesn't look to be called at the protocol side... (expect if I've missed it)
Btw, even if the token contract is Pausable, we would not be able to pause it (not our own), so...

contracts/DAITokenWrapper.sol
```diff
@@ -0,0 +1,150 @@
...
+    }
+}
+
+/**
+ * @title IDAI
+ * @notice Interface for the purpose of calling the permit function on the deployed DAI token
+ */
+interface IDAI {
+ 
+    function name() external pure returns (string memory);
```
levalleux-ludo, line 138: I think we don't need to define the name() property because we d'ont call it

contracts/DAITokenWrapper.sol
```diff
@@ -0,0 +1,150 @@
...
+    
+    function permit(address holder, address spender, uint256 nonce, uint256 expiry,
+                    bool allowed, uint8 v, bytes32 r, bytes32 s) external;
+
+    function transferFrom(
```
levalleux-ludo, line 143: I think we don't need to define the transferFrom() method because we don't call it

levalleux-ludo (commented):
Please see my comments (more generally some questions)

contracts/DAITokenWrapper.sol
hswopeams: Good point. I'll keep the set function as well. It's not likely on Ethereum mainnet, but I have noticed that on Polygon, there are multiple DAI child contracts, even on Polygon mainnet. So it's possible we could set the wrong address at construction time and need to reset it.

contracts/DAITokenWrapper.sol
hswopeams: The name function is need for unit testing. It's called by the function that creates the digest.

contracts/DAITokenWrapper.sol
hswopeams: It's needed for unit testing. If it's not defined in the interface, it can't be used for the mock. The mock needs to return a value for the transferFrom, or the unit tests won't work. It gets set like this: ` await mockDAI.mock.transferFrom`

contracts/DAITokenWrapper.sol
hswopeams: I suppose it's not really needed. I'll remove Pausable.

contracts/DAITokenWrapper.sol
hswopeams: This will be removed when the `unpause` function is removed

hswopeams: Pauable removed

contracts/DAITokenWrapper.sol
hswopeams: Removed

contracts/DAITokenWrapper.sol
levalleux-ludo: OK

contracts/DAITokenWrapper.sol
levalleux-ludo: OK

contracts/DAITokenWrapper.sol
levalleux-ludo: OK

levalleux-ludo (dismissed):

mischat (approved):
This looks good to me, 

I would make sure that we have more comments in the code, which we can do after merging this code. 

I also feel like we could refactor the code to have less code repetition.

as an aside, i feel like we should make sure that our PRs are single purpose. There is a lot of noise in this PR which is related to the rename of one of the contracts.

scripts/contracts-hardhat.json
```diff
@@ -1,8 +1,8 @@
 {
-  "network": "hardhat",
-  "flo": "0x6DfC760cf9BBaE3Bd4b6DfCE536aa372E9A611a9",
-  "erc1155erc721": "0x5958E222Cf51a94Abc1228a3d709bbCeD9f660b7",
-  "voucherKernel": "0x4FF0B67c6eE79Ed6d739419D70c7299b7285b3Cc",
-  "cashier": "0x39C1c6bFc6f1EbBD4e45C11F44099066B919E577",
-  "br": "0x1507e4D16E01981f6A3Da847C9367DE0A40a6671"
+  "network": "localhost",
```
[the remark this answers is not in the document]
hswopeams: Yes. Localhost is when you run hardhat as a separate node, for instance when forking

test/2_test_fullpath_with_permit.ts
```diff
@@ -145,15 +140,26 @@ describe('Cashier and VoucherKernel', () => {
...
     );
-    await contractFundLimitsOracle.setTokenLimit(
+    await contractTokenRegistry.setTokenLimit(
       contractBSNTokenDeposit.address,
       constants.TOKEN_LIMIT
     );
-    await contractFundLimitsOracle.setETHLimit(constants.ETHER_LIMIT);
+    await contractTokenRegistry.setETHLimit(constants.ETHER_LIMIT);
+
+    //Set Boson Token as it's own wrapper so that the same interface can be called in the code
```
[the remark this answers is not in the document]
hswopeams: I don't know the history of why this limit is set. I only know how it works from a technical point of view.

contracts/DAITokenWrapper.sol
```diff
@@ -0,0 +1,115 @@
...
+    {
+        return daiTokenAddress;
+    }
+}
+
+/**
+ * @title IDAI
+ * @notice Interface for the purpose of calling the permit function on the deployed DAI token
+ */
+interface IDAI {
```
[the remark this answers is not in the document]
hswopeams: Yes, but those interfaces are full interfaces for a whole contract. This one is only ever meant to be called by the wrapper and isn't a complete interface, so it seemed strange to put it in the interfaces directory.

contracts/DAITokenWrapper.sol
```diff
@@ -0,0 +1,115 @@
...
+        external
+        override
+        notZeroAddress(owner)
+        notZeroAddress(spender)
+    {
+        require(deadline == 0 || block.timestamp <= deadline, "PERMIT_EXPIRED");
+        require(r != bytes32(0) && s != bytes32(0), "INVALID_SIGNATURE_COMPONENTS");
+        uint nonce =  IDAI(daiTokenAddress).nonces(owner);
+        IDAI(daiTokenAddress).permit(owner, spender, nonce, deadline, true, v, r, s);
+        emit LogPermitCalledOnToken(daiTokenAddress, owner, spender, 0);
```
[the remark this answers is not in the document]
hswopeams: The linter must not have picked that up.

zajck (approved):

PR #146, opened 2021-09-28
Function parameters with underscore

bidyut-arianelabs:
This PR adds _ to variable names in function parameters.


zajck (changes requested):
Modifiers with parameters are also not consistent.
Please, fix also those.

zajck (dismissed):

hswopeams (changes requested):
- ERC1155NonTransferable has underscore **after** the parameter name in `constructor(string memory uri_)`

hswopeams (approved):

zajck (approved):

PR #149, opened 2021-09-29
Check value of ERC-20 transfers in BosonRouter

### LINKED ISSUE #81: Check value of ERC-20 transfers in BosonRouter
Minor volatile code issue: `BosonRouter.sol` is not checking for return value of `transfer()` and `transferFrom()` of the ERC-20.
Reason: the contract has reached max size and needs refactoring.

Recommendation: user OpenZeppelin's `SafeERC20.sol`.

bidyut-arianelabs:
This PR implements `safeTransferFrom` instead of `transferFrom` in `BosonRouter.sol`.


hswopeams (changes requested):
UsingHelpers.sol is also a utility, but we don't have it in a separate folder. This is the only library we have, so it might make sense to put it in a different folder, but then I would call the folder `libs`. This is because the aforementioned UsingHelpers.sol is also a utility but not a library.
@zajck What do you think?

@bidyut-arianelabs Does the addition of this library necessitate changes to the deploy.ts?

zajck:
> UsingHelpers.sol is also a utility, but we don't have it in a separate folder. This is the only library we have, so it might make sense to put it in a different folder, but then I would call the folder `libs`. This is because the aforementioned UsingHelpers.sol is also a utility but not a library. @zajck What do you think?
> 
> @bidyut-arianelabs Does the addition of this library necessitate changes to the deploy.ts?

I agree, let's put it in separate folder (which is alreday done at the time of writting of this comment).

contracts/libs/SafeERC20WithPermit.sol
```diff
@@ -0,0 +1,66 @@
...
+ * @title SafeERC20WithPermit
+ * @dev Wrappers around ERC20 operations that throw on failure (when the token
+ * contract returns false). Tokens that return no value (and instead revert or
+ * throw on failure) are also supported, non-reverting calls are assumed to be
+ * successful.
+ * To use this library you can add a `using SafeERC20WithPermit for IERC20WithPermit;` statement to your contract,
+ * which allows you to call the safe operations as `token.safeTransfer(...)`, etc.
+ */
+library SafeERC20WithPermit {
+    using SafeMath for uint256;
```
zajck, line 19: Does SafeMath have any effect here? I think not and it should be removed.

zajck (commented):
I think we should also unit test the code. It does look ok, does not break any old code, but just to make sure it does what it is supppose to.

hswopeams (changes requested):
The library needs to be linked to the contract. This needs to be done in unit tests (which deploy their own contracts) and in deploy.ts. Verification of the lib should be added to verify.ts as well

https://hardhat.org/plugins/nomiclabs-hardhat-ethers.html#library-linking

Let me know if you need a walkthrough of deploy.ts and verify.ts

Also please update MockBosonRouter.sol. After updating, make sure 5 is passed into `createPaymentMethod` on line 170

contracts/libs/SafeERC20WithPermit.sol
bidyut-arianelabs: Removed.

test/1_test_fullpath.ts
```diff
@@ -1207,6 +1219,38 @@ describe('Voucher tests', () => {
       assert.isTrue(sellerBalanceAfter.eq(expectedSellerBalance));
     });
   });
+
+  describe('TransferFrom: It is safe to interact with older ERC20 tokens', function () {
+    let sellerInstance;
+    beforeEach('set seller as boson router', async () => {
```
hswopeams, line 1225: Is this supposed to say Set **mock** as Boson Router?

bidyut-arianelabs: `Set mock as Boson router` makes more sense. I will update.

scripts/deploy.ts
```diff
@@ -24,6 +24,7 @@ class DeploymentExecutor {
   dai_token;
   gate;
   erc1155NonTransferable;
+  safeERC20WithPermit;
```
zajck, line 28: Do we actually need to deploy the library? Since it has no external methods, a think compiler just takes them and there is no nee for separate library instance.
Even looking further down the code (and in tests), I don't see that you would need to link BosonRouter and this library (as is needed when library has external methods). Or I am missing something?

bidyut-arianelabs: We need to deploy the library so that it can give us the deployed address.
We will use that deployed address to verify the library source code using the `verify.ts` in etherscan.

contracts/BosonRouter.sol
```diff
@@ -757,7 +759,8 @@ contract BosonRouter is
     function transferFromAndAddEscrow(address _tokenAddress, uint256 _amount)
         internal
     {
-        IERC20WithPermit(_tokenAddress).transferFrom(
+        SafeERC20WithPermit.safeTransferFrom(
+            IERC20WithPermit(_tokenAddress),
             msg.sender,
             address(cashierAddress),
             _amount
```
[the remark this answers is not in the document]
bidyut-arianelabs: removing `using SafeERC20WithPermit for IERC20WithPermit;` in next commit.

scripts/deploy.ts
zajck: Yes, but why? If we deploy it and verify it, but our boson router is not actually linked to it, what's the purpose then?

I'm pretty sure that the library will be verified as part of the boson router verification and not as standalone.
Similar as for uniswapV3
https://etherscan.io/address/0xe592427a0aece92de3edee1f18e0157c05861564#code

hswopeams: I was going by this: https://docs.soliditylang.org/en/v0.7.6/using-the-compiler.html?highlight=library%20linking#library-linking

And also this: https://docs.soliditylang.org/en/v0.7.6/contracts.html?highlight=library#libraries
specifically: "Libraries are similar to contracts, but their purpose is that they are deployed only once at a specific address". Since we have created the library ourselves and are not pointing to and already deployed OpenZeppelin library, it would need to be deployed at some point.
But since the functions are` internal`, you might be right.

bidyut-arianelabs: ok, will remove updates from `deploy.ts` and `verify.ts` in the next commit.

hswopeams (approved):

zajck (approved):

PR #153, opened 2021-10-01
Add slither to build

bidyut-arianelabs:
This PR adds `slither` static analysis tool into the build.

# Dependency:
Docker must be installed on your Mac / Linux.

# How to run slither.
1. You can run it when creating a build as `./go build`.
2. Alternatively, You can also run it as `./go contracts:slither`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ At a minimum, you'll need:
 * Bundler (> 2)
 * Git
 * direnv
+* Docker
 
 For instructions on how to get set up with these specific versions:
 * See the [OS X guide](docs/setup/osx.md) if you are on a Mac.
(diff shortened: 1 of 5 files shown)


bidyut-arianelabs:
FYI, Build fails because CI server does not have `Docker` installed and running.

cliffhall:
Are we over this yet?

cliffhall:
@mischat We've reached the point of diminishing returns on this. We can investigate it for v2 if you still feel strongly about getting it into the build process, but it's not going to help us on this project. Adding it to the build will only extend the build time, and we've determined that it is very slithery to pin down, with regard to its output and filtering. It can still be run locally if need be.

PR #158, opened 2021-10-06
Add zero address checks

hswopeams:


contracts/VoucherKernel.sol
```diff
@@ -172,11 +172,19 @@ contract VoucherKernel is IVoucherKernel, Ownable, Pausable, UsingHelpers {
...
+
     /**
      * @notice Construct and initialze the contract. Iniialises associated contract addresses, the complain period, and the cancel or fault period
      * @param _tokensContract address of the associated ERC1155ERC721 contract instance
      */
-    constructor(address _tokensContract) {
+    constructor(address _tokensContract)
+    notZeroAddress(_tokensContract)
+    {
+        require(_tokensContract != address(0), "0A");
```
bidyut-arianelabs, line 187: Shouldn't this line be removed now that we have `notZeroAddress` modifier?

bidyut-arianelabs (changes requested):

contracts/VoucherKernel.sol
hswopeams: Yep. I forgot to remove it. I'll remove it in the next commit.

bidyut-arianelabs (dismissed):
Looks good to me 👍

zajck (commented):
Generally it looks ok, but I have one question before approving:

In some functions (e.g. `Cashier._withdrawPayments`, `Cashier._withdrawDeposits` and some other places) , we do check `require(_recipient != address(0), "UNSPECIFIED_ADDRESS");` which is basically the same as it would be done with notZeroAddress modifier. I know that this is there already from before, but I just wonder if there is any specific reason not to use the modifier?

contracts/Cashier.sol
```diff
@@ -815,7 +819,9 @@ contract Cashier is ICashier, UsingHelpers, ReentrancyGuard, Ownable, Pausable {
         uint256 _amount,
         uint8 _paymentMethod,
         uint256 _tokenIdSupply
-    ) internal {
+    ) internal    
+      notZeroAddress(_recipient)
+    {
         require(_recipient != address(0), "UNSPECIFIED_ADDRESS");
```
zajck, line 825: `require` can now be removed`require` can now be removed

zajck (changes requested):
Beside what I put the in code comments - there are other places that we use two different checks for `nonZeroAddress` (in some functions with modifiers, in others with explicit require statements). This happens in `Cashier`, `Gate` and `VoucherKernel`. But changing this might even be a task for separate PR.
We can make it a separate issue and I'll aprove this after you remove those two lines I commented.

bidyut-arianelabs (approved):
Looks good to me 👍

zajck (approved):

PR #163, opened 2021-10-08
Slither extternal calls and shadowing

zajck:
This PR resolves warnings about
- functions declared public instead of external
- parameters that shadow other parameters

### Prerequiste for testing:
```
Slither must be installed.
```

### Steps for testing this PR:
```
    1. Run slither on main using command `slither . --filter-paths "openzeppelin"`
    2. Then run slither in my branch slither-ext-calls-shadows
    3. Compare results of 1 and 2.
```


bidyut-arianelabs (approved):
Looks good to me 👍

hswopeams (approved):

PR #164, opened 2021-10-08
Silence two slither issues

hswopeams:
Mint return value: If it doesn't revert, this mint function overload returns a hardcoded true value.
Thus. the return value isn't necessary

Initialize local variable:
boolean variables are automatically initialized to false


zajck (dismissed):

bidyut-arianelabs (dismissed):
Looks good to me 👍

bidyut-arianelabs (approved):
Looks good to me 👍

zajck (approved):

PR #165, opened 2021-10-11
Unit tests - 2_test_fullpath_with_permit

zajck:
Unit test `2_test_fullpath_with_permit is updated`, to more prcisely check emited events and state changes.

Utils functions were changes, so I had to update some other test (`3_withdrawals`, `4_pausing_contracts`, `6_erc1155721`, `11_gate` and `12_conditional_commit`) files too in order to build to success. Otherwise these files were not changed.

Some other comments:
- did not add tests for `commit`, `redeem`, `refund` and `COF` for every combination, since these methods are properly tested in `1_test_fullpath` and are not payment type specific
- some block of the code could be rewritten so the style would be more consistent, but at the moment it would just take time without any other positive effect


bidyut-arianelabs (changes requested):

hswopeams (changes requested):

hswopeams:
It looks good overall. Much improved. There are still a few things we need to address.

zajck (commented):
All inline code comments are resolved.

Not resolved yet:
`IERC721Receiver(_to).onERC721Received.selector` and `IERC1155Receiver(_to).onERC1155Received.selector`

bidyut-arianelabs (changes requested):

bidyut-arianelabs (approved):
Looks good to me 👍

hswopeams (approved):

PR #166, opened 2021-10-11
Gate register voucher

zajck:
Same gate contract now allows to `registerVoucherSetId` either from BosonRouter or directly by Owner.

MockGate was removed from repository since all corresponding test cases are now done using Gate.


hswopeams (approved):

bidyut-arianelabs (approved):
Looks good to me 👍

PR #168, opened 2021-10-13
This is the metatx relayer contract, with tests

mischat:
This should be features complete.


zajck (commented):
Two things:

1. Is it desired, than when functions are called via metaTx, that the "operator" field in events is contract address? The actual sender is owner() and I think it should be reflected here. Since contract is already using Context, we can override `_msgSender()` to return `owner()` when it's called by itself
2. Only method `mint` is tested for metatx, but there are plenty more that should be checked (mintBatch, burn, burnBatch, pause, unpause).

bidyut-arianelabs (approved):
Looks good to me 👍

hswopeams:
Looks good.

hswopeams (approved):

PR #170, opened 2021-10-13
We still have issues with the package-lock.json ... not sure what is

mischat:
going on here ...


zajck (approved):

mischat:
I guess is the question is, do we want it to have "false" or to resolve to a url, if the later, then we should keep the changes in the PR...

bidyut-arianelabs (approved):
`resolved` should not be `false`. Hence approving this PR.

PR #172, opened 2021-10-13
Update deploy.ts to set metadata URIs correctly

hswopeams:
Updated deploy.ts to set metadata URIs for ERC1155NonTransferable and ERC1155ERC721 using environment variables. Also updated .env.example. This enables the deployer to use different URIs per environment by creating different .env files. '.env.*' was added to git ignore so that people can create .env.test, .env.staging, .env.prod, etc. without these getting accidentally checked into GitHub.


zajck (changes requested):

zajck (dismissed):

bidyut-arianelabs (changes requested):

levalleux-ludo (dismissed):

bidyut-arianelabs (dismissed):
Looks good to me 👍

bidyut-arianelabs (approved):

levalleux-ludo (approved):

PR #173, opened 2021-10-13
Bring 6_erc1155721.ts up to current standards

bidyut-arianelabs:
In this PR, 
1. I am adding new unit test cases for the `ERC1155ERC721.sol` contracts.
2. Adding both happy and negative unit test cases.


zajck (changes requested):
There exists two function for safeTrasferFrom for token 721:
```
    function safeTransferFrom(
        address _from,
        address _to,
        uint256 _tokenId
    )
```

and

```
    function safeTransferFrom(
        address _from,
        address _to,
        uint256 _tokenId,
        bytes memory _data
    )
```

In tests we always test only that with 4 parameters.
I think that's completely fine and not further action is needed. It's just that coverage report will always point that out.

And another thing from coverage report - `supportsInterface` is not tested. Again it's not a big deal, but just adding few test for that should be quick and straigth forward.

hswopeams (changes requested):
A few general comments:

- The number and type of test cases are good. Klemen and I noted a few places where we'd like to see some test cases added.
- In the positive test cases where tokens are minted, burned, or transferred only the events are checked in most places. The state also needs to be checked. This means checking the `balanceOf` and `ownerOf` (where appropriate) in the positive test cases

zajck:
All comments were resolved well.

Just for the sake of completeness, I think you should also add the following tests

```
ERC1155
[safeTransfer1155] Should be able to safely transfer to EOA
[safeBatchTransfer1155] Should be able to safely batch transfer to EOA
[safeBatchTransfer1155] Should be able to safely batch transfer to contracts that support ERC1155
[NEGATIVE][mint] Should revert when to is a contract that cannot receive it
[NEGATIVE][mintBatch] Should revert when to is a contract that cannot receive it
ERC721
[NEGATIVE][mint] Should revert when to is a zero address
[NEGATIVE][mint] Should revert when to is a contract that cannot receive it
```

hswopeams (changes requested):

zajck (changes requested):

hswopeams (changes requested):
It looks really good. The only request I have remaining is to add some comments regarding the use of setvoucherKernelAddress to set the deployer address as the VoucherKernel address. See comments above.

hswopeams (approved):

zajck (approved):

PR #197, opened 2021-10-15
Unit tests 3 withdraw scenarios

zajck:
Updated section `'Withdraw scenarios'`

Other two sections were not updated (or removed). If/when they are moved to separate file, they should be removed from here.

PR #203, opened 2021-10-18
Unit test 3 'Seller cancels uncommitted voucher set'

zajck:
Updated section `Seller cancels uncommitted voucher set`

I'll leave it as a draft, until #197 is approved and merged, and open ti to review afterwards.


hswopeams (changes requested):
Please add `TransferSingle` check (See previous comment).

In addition, I don't see any tests of the `Cashier.withdrawDepositsSe` function. It should only be callable from the BosonRouter (negative test). I also notice that we are not doing a check on the `_messageSender` paramater to check that address is the actual owner of the voucher set. A check is being done in` VoucherKernel.cancelOrFaultVoucherSet`, but what if the code were ever changed in some way such that `require(getSupplyHolder(_tokenIdSupply) == _issuer, "UNAUTHORIZED_COF");` was (accidentally) removed from ` VoucherKernel.cancelOrFaultVoucherSet`?  `VoucherKernel.getSupplyHolder` can be used to check the voucher set owner

hswopeams (approved):

bidyut-arianelabs (approved):
Looks good to me 👍

PR #205, opened 2021-10-19
Unit tests for 3_withdrawals.ts: Withdraw on disaster

bidyut-arianelabs:
In this PR, I am adding the remaining unit test cases for the `Withdraw on Disaster` scenarios.

Depends on https://github.com/bosonprotocol/contracts/pull/197


hswopeams (changes requested):
It looks pretty good. I have a a few small requests for change. In addition, I have a functonal question. According to the `setDisasterState` comments, "Once this functions is triggered, contracts cannot be unpaused anymore".

I don't see a test case for this. There's also no indication that the functionality was built. I think the question is if it should be added or not. I think it's a functional question we shoud discuss in the team.

zajck (changes requested):

zajck (dismissed):

zajck (approved):

hswopeams (approved):

PR #206, opened 2021-10-19
Publish npm package

dohaki:
This adds a GitHub workflow to publish the contracts and relevant files as a package to npm. The workflow is triggered if a new release is created. The version number of the package gets taken from the git tag. For older releases, we would need to publish manually though.

Would appreciate some feedback on the package name `@bosonprotocol/contracts` and included files if they make sense. The only thing that bothers me with the package name `@bosonprotocol/contracts` is that you would have to import like this
```js
// In Solidity
import "@bosonprotocol/contracts/contracts/interfaces/IBosonRouter.sol"
// In JS
import ABI from "@bosonprotocol/contracts/abi/contracts/BosonRouter.json"
```
Maybe a package name like `@bosoprotocol/core` is better? The you would not have the word `contracts` doubled in every import:
```js
// In Solidity
import "@bosonprotocol/core/contracts/interfaces/IBosonRouter.sol"
// In JS
import ABI from "@bosonprotocol/core/abi/contracts/BosonRouter.json"
```

**TODO: Add the env var `NPM_TOKEN` with respective privileges to `Settings > Secrets`.**

### DIFF
--- a/.github/workflows/publish-pre-release.yaml
+++ b/.github/workflows/publish-pre-release.yaml
@@ -1,7 +1,7 @@
-name: Publish to npm
+name: Publish pre-release to npm
 on:
   release:
-    types: [created]
+    types: [prereleased]
 jobs:
   publish:
     runs-on: ubuntu-latest
@@ -19,4 +19,6 @@ jobs:
       - uses: JS-DevTools/npm-publish@v1
         with:
           token: ${{ secrets.NPM_TOKEN }}
-          access: "public"
+          access: public
+          tag: beta
+          check-version: false
--- a/.github/workflows/publish-pre-release.yaml
+++ b/.github/workflows/publish-pre-release.yaml
@@ -15,7 +15,7 @@ jobs:
           node-version: "12.x"
           registry-url: "https://registry.npmjs.org"
       - run: npm install
-      - run: npm --no-git-tag-version version "${{ env.RELEASE_VERSION }}"
+      - run: npm --allow-same-version --no-git-tag-version version "${{ env.RELEASE_VERSION }}"
       - uses: JS-DevTools/npm-publish@v1
         with:
           token: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/publish-release.yaml
+++ b/.github/workflows/publish-release.yaml
@@ -1,0 +1,24 @@
+name: Publish release to npm
+on:
+  release:
+    types: [released]
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set env RELEASE_VERSION
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+      # Setup .npmrc file to publish to npm
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "12.x"
+          registry-url: "https://registry.npmjs.org"
+      - run: npm install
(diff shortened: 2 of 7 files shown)


damir-bajramovic (changes requested):
I am not sure what is the intended way of using this package. Is it like the contracts from `OpenZeppelin`?

I don't see an `.npmignore` file which is kinda important. If that file is missing, the `.gitignore` file is used by default.

Do we want to expose the ABIs with this package? If so, we need to push them to the npm registry by excluding them from the `.npmignore` file. 

Also, I believe that ABIs are used in a bunch of different places. It would be nice to publish them separately to the NPM registry, so we can pull them without pulling the rest of the code from this repository. 
We could automate the process when pushing new versions of the contracts to the `main` branch or something. 
Something similar can be witnessed in the `ethers` repository. They are using multiple packages as dependencies in one repository. Feel free to explore the structure.

damir-bajramovic (dismissed):

hswopeams (commented):
Is there a way to test this to make sure it only publishes releases and not pre-release tags?

hswopeams:
> @hswopeams
> 
> > Is there a way to test this to make sure it only publishes releases and not pre-release tags?
> 
> I tested it on a fork https://github.com/dohaki/bosonprotocol-contracts. But the workflow currently triggers on every release AND pre-release created. See
> 
> https://github.com/bosonprotocol/contracts/blob/fdca3b65b73dcb37addf593752ebdda2b196dd5d/.github/workflows/publish.yaml#L2-L4
> 
> Here are the events which can be used https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release. But I think it makes sense to also publish pre-releases. I mean the version suffix `-rc.X` implies that the package is a pre-release. And the main reason why I added this was because we had to manually update the ABIs in multiple places for testing the new contracts. If we would only publish releases then we would still need to do this.
> 
> What I could do though is to add a separate publish workflow. So we would have
> 
> 1. `Publish Pre-Releases`
> 
> * Triggered for pre-releases only
> * Uses [npm dist tag](https://docs.npmjs.com/adding-dist-tags-to-packages) `next` or similar
> 
> 1. `Publish Releases`
> 
> * Triggered for releases only
> * Uses default npm dist tag `latest`

I don't agree we should push pre-releases. There's a reason they're not releases. We will discuss this in the protocol team and make a decision.

dohaki:
@hswopeams 
> I don't agree we should push pre-releases. There's a reason they're not releases. We will discuss this in the protocol team and make a decision.

Ok, thanks. Let me know how I should proceed. Just to be clear: with pre-release, I am referring to https://github.com/bosonprotocol/contracts/releases/tag/v1.0.0-rc.1 for example. I don't see a reason why we would have `v1.0.0-rc.1` published on GitHub but not on npm 🤔 

> The npm registry should NOT be updated every time we merge with main. Every main merge is NOT a release.

That is also not what I intended. As I said, the workflow is only triggered if you create a release or pre-release. So basically it would have been triggered for every listed release/pre-release you can find here https://github.com/bosonprotocol/contracts/releases and not on every merge to main.

mischat (approved):
This looks good to me. I like the idea of building into GHA and streamlining our contract integrations is a must for us.

mischat (changes requested):
Is there nothing we can do about the /contracts/contracts ... can we not configure the job to take all of the files from the ./contracts folder ?

mischat (dismissed):
This is as discussed and as expected; this looks good to me.

damir-bajramovic (dismissed):

mischat (approved):
LGTM!

PR #212, opened 2021-10-20
Remediate BRR-04C

### LINKED ISSUE #181: BRR-04C: Potential Indexing Optimization
## BRR-04C: Potential Indexing Optimization

| [Code Style](/appendix/finding-types#code-style) | <span class="o-severity o-informational">Informational</span> | [BosonRouter.sol:L48-L53](https://github.com/bosonprotocol/contracts/blob/817b6814e4e93939fe237c525107f23ad67ce2da/contracts/BosonRouter.sol#L48-L53) |

### Description:

The `LogOrderCreated` event specifies the `_tokenIdSupply` as `indexed` whilst it is meant to change on each emittance.

### Example:

```solidity
event LogOrderCreated(
    uint256 indexed _tokenIdSupply,
    address _seller,
    uint256 _quantity,
    uint8 _paymentType
);
```

### Recommendation:

We advise the `indexed` keyword to be dropped from it as the keyword is meant to be used on values that are expected to be equal between invocations and can thus be aggregated. Alternatively, we advise the `_seller` to become `indexed`.

## Remediation Strategy

- [x] Follow the recommendation to add the indexed keyword to _seller parameter.
- [x] #211

The underscore idiom for method arguments is so that within the method body, we can easily distinguish local and state variables from method arguments. However only clients of the contract will read the parameters of events, and the underscore has no meaning whatsoever in that context, making access of those parameters needlessly awkward in client code.

cliffhall:
* In BosonRouter.sol,
  - LogOrderCreated event
    - added indexed keyword to _seller argument
  - Verified compilation and unit tests all succeed.

* This fixes #181


zajck (approved):

hswopeams (changes requested):
The recommended change includes removing the `indexed` keyword from _tokenIdSupply

hswopeams (approved):

Issue #202, opened 2021-10-17
Unnecessary Require statements


mudgen:
https://github.com/bosonprotocol/contracts/blob/043d9304860ad0722eeef810d04a33418f9594c7/contracts/VoucherKernel.sol#L303-L310

The require statements here are not necessary except for the one about quantity.  But I think the check for quantity should be placed in the `createTokenSupplyId` function instead.

The `createOrder` function is only called in the `createTokenSupplyId` function and no where else.  The `createTokenSupplyId` function makes these requires unnecessary.

PR #223, opened 2021-10-22
Code coverage

zajck:
With this PR we are at 100% coverage for almost all contracts.

This is the report:

[coverage.zip](https://github.com/bosonprotocol/contracts/files/7399825/coverage.zip)

The reason why it's still not there are the following:

## Latest PRs

Some code that was recently merged into main, does not cover all branches, therefore branches for `Cashier.sol` and `C1155ERC721.sol` are not 100%

### createTokenSupplyId

https://github.com/bosonprotocol/contracts/blob/2b5bb39c772b618427a738f67b9b2174d1f412f9/contracts/VoucherKernel.sol#L238-L248

The way key is created it should have negligible possibility that it is used already.
Cannot test the require statement.

### createOrder

Require in lines 314 and 315 will never fail.

https://github.com/bosonprotocol/contracts/blob/2b5bb39c772b618427a738f67b9b2174d1f412f9/contracts/VoucherKernel.sol#L309-L316

`createOrder` is private and called only from `createTokenSupplyId`, so we cannot test differen input parameters
Described also in https://github.com/bosonprotocol/contracts/issues/202

suggestion: follow the recommendation from the issue and remove it

### burnSupplyOnPause
https://github.com/bosonprotocol/contracts/blob/2b5bb39c772b618427a738f67b9b2174d1f412f9/contracts/VoucherKernel.sol#L457-L463

VoucherKernel.burnSupplyOnPause has modifier `onlyFromCashier`, yet it's never actually called in cashier.
I wrote tests for that nontheless, question is only what is its purpose.

Suggestion: investigate what is the intended use of this method.

### generateTokenType
https://github.com/bosonprotocol/contracts/blob/2b5bb39c772b618427a738f67b9b2174d1f412f9/contracts/VoucherKernel.sol#L470-L480

`voucherKernel.generateTokenType` is internal and called only once with input parameter `true`
Else path is unreachable

Suggestion: remove input parameter and conditional block

### complain
[voucherKernel: 601-606](https://github.com/bosonprotocol/contracts/blob/2b5bb39c772b618427a738f67b9b2174d1f412f9/contracts/VoucherKernel.sol#L601-L606)

We are inside a block where `isStateRedemptionSigned(vouchersStatus[_tokenIdVoucher].status) || isStateRefunded(vouchersStatus[_tokenIdVoucher].status)` is `true`
If we wanted to reach line 601, `!isStatus(vouchersStatus[_tokenIdVoucher].status, IDX_CANCEL_FAULT)` should be false, which happens when voucher is in `IDX_CANCEL_FAULT` state. However, if voucher is in `IDX_CANCEL_FAULT` state, then `isStateRedemptionSigned(vouchersStatus[_tokenIdVoucher].status) || isStateRefunded(vouchersStatus[_tokenIdVoucher].status)` is `false` and we are not even in the block starting at 588.
And in this case code goes to block [648-663](https://github.com/bosonprotocol/contracts/blob/2b5bb39c772b618427a738f67b9b2174d1f412f9/contracts/VoucherKernel.sol#L648-L663) where actually the exact same check is done, that should be done in block starting at 601.

Suggestion: remove unnecessary conditional block.

[voucherKernel: 631-636](https://github.com/bosonprotocol/contracts/blob/2b5bb39c772b618427a738f67b9b2174d1f412f9/contracts/VoucherKernel.sol#L631-L636)

We are inside a block where `isStateExpired(vouchersStatus[_tokenIdVoucher].status` is `true`.
To reach line 632 `!isStatus(vouchersStatus[_tokenIdVoucher].status, IDX_CANCEL_FAULT)` should be false, which happens when voucher is in `IDX_CANCEL_FAULT` state. However, if voucher is in `IDX_CANCEL_FAULT` state, then isStateExpired(vouchersStatus[_tokenIdVoucher].status)` is `false` and we are not even in the block starting at 620.
And in this case code goes to lin 648. But now, contrary to previous case, checks are different. Block [631-636](https://github.com/bosonprotocol/contracts/blob/2b5bb39c772b618427a738f67b9b2174d1f412f9/contracts/VoucherKernel.sol#L631-L636) compare block.timestamp to tPromise.validTo + complainPeriod, while block [648-663](https://github.com/bosonprotocol/contracts/blob/2b5bb39c772b618427a738f67b9b2174d1f412f9/contracts/VoucherKernel.sol#L648-L663)  compare it to vouchersStatus[_tokenIdVoucher].complainPeriodStart + complainPeriod

Suggestion: investigate further, what is the correct behaviour 

### _withdrawPayments

`require` in `_withdrawPayments` will never fail.
https://github.com/bosonprotocol/contracts/blob/c2fa335cc3e22b8154314646d68a7fc79ccd4eac/contracts/Cashier.sol#L787

`_withdrawPayments` is called only from `withdraw` and only if `voucherDetails.price2issuer > 0` or `voucherDetails.price2holder > 0` https://github.com/bosonprotocol/contracts/blob/c2fa335cc3e22b8154314646d68a7fc79ccd4eac/contracts/Cashier.sol#L233-L240 https://github.com/bosonprotocol/contracts/blob/c2fa335cc3e22b8154314646d68a7fc79ccd4eac/contracts/Cashier.sol#L251-L259

Since `price2issuer` and `price2holder` are input parameters for `amount` of `_withdrawPayments`,  `require` will always be true.

Suggestion: remove unecessary require statement.

# Other comments

## Tests that might get deprecated

In some contracts (Cashier, VoucherKernel) we still check every time, when calls are made from Cashier or VoucherKernel that addresses for that contracts are actually set.
If we followed the pattern we have in other contracts, where those variables are set in constructor, we could remove the following tests:

13_cashier: [NEGATIVE] Should revert if boson router is not set
14_voucherKernel: [NEGATIVE] Should revert if boson router is not set
14_voucherKernel: [NEGATIVE] Should revert if cashier address is not set

Moreover, we could remove `RevertReasons.UNSET_ROUTER` and adjust `deployContracts` functions in files 13_cashier, 14_voucherKernel to be without input parameteres, which tells if contractas should be set or not

## Cashier - VoucherKernel setter

In Cashier we can set VoucherKernel only in constructor. We do not have a separate setter, to change it in case something went wrong.

## Two very similar checks
`voucherKernel.checkOrderFillable` does check if `_to` is contract that it can receive erc721 tokens

https://github.com/bosonprotocol/contracts/blob/2b5bb39c772b618427a738f67b9b2174d1f412f9/contracts/VoucherKernel.sol#L382-L388

Then similar check is done in `voucherKernel.extract721`

https://github.com/bosonprotocol/contracts/blob/2b5bb39c772b618427a738f67b9b2174d1f412f9/contracts/VoucherKernel.sol#L417-L428

checkOrderFillable and extract721 are always called only from one method (fillOrder) so we could move one of this check there and get rid of the other.

## Code changes
I made some small solidity code changes.

`voucherStatus.checkOrderFillable`
- checking if holder is zero address already with modifier (before it was checked only after checking if it's contract or not)

`voucherStatus.complain`
- switched the order of `ALREADY_COMPLAINED`, `ALREADY_FINALIZED`. More logical -> Imagine a scenario where was a complaint and it was then finalized. If you call another complain afterwards it make more sense to get error `ALREADY_FINALIZED` than `ALREADY_COMPLAINED`, since it's more recent state.

`voucherStatus.cancelOrFault`
- similar as in voucherStatus.complain

## Soft unhappy paths

Some functions can be called in some scenarios at some point, where they do not do anything - they do not emit any event or change state, but they also do not revert. This happens because they contain if statements, that allow gracefully execution (i.e. else path does not exist)
This happens in `withdraw`, `triggerExpiration` and `finalize`.
I think this is not good for two reasons -> whoever submits a transaction and sees that tx suceeded, would expect some actual changes. And secondly, testing is harder, since we cannot rely on revert reasons but must check that events were not emitted and state remained the same.

Suggestion: where possible, throw error, if there are no state changes.


bidyut-arianelabs (changes requested):
Looks good. Few small changes needed.

hswopeams (commented):
It looks pretty good. I have a few questions and comments.

hswopeams (dismissed):

bidyut-arianelabs (dismissed):
Looks good to me 👍

cliffhall (dismissed):
LGTM! 👍
I've added [my comments](https://github.com/bosonprotocol/contracts/pull/223#issuecomment-951323338) on your outstanding questions in the discussion.

hswopeams (approved):

PR #233, opened 2021-10-26
CAS-07C: Variable Mutability Specifier

### LINKED ISSUE #191: CAS-07C: Variable Mutability Specifier
## CAS-07C: Variable Mutability Specifier

| [Gas Optimization](/appendix/finding-types#gas-optimization) | <span class="o-severity o-informational">Informational</span> | [Cashier.sol:L25](https://github.com/bosonprotocol/contracts/blob/817b6814e4e93939fe237c525107f23ad67ce2da/contracts/Cashier.sol#L25), [L83](https://github.com/bosonprotocol/contracts/blob/817b6814e4e93939fe237c525107f23ad67ce2da/contracts/Cashier.sol#L83) |

### Description:

The linked variable is assigned to only once during the contract's `constructor`.

### Example:

```solidity
/**
 * @notice Construct and initialze the contract. Iniialises associated contract addresses. Iniialises disaster state to false.    
 * @param _voucherKernel address of the associated VocherKernal contract instance
 */
constructor(address _voucherKernel) {
    voucherKernel = _voucherKernel;
    disasterState = false;
}
```

### Recommendation:

We advise it to be set as `immutable` greatly optimizing its read access gas cost.

## Remediation Strategy

The contract whose address is set only in the constructor may later be superseded, should vulnerabilities be found and new versions deployed. Rather than following the recommendation to add immutable to the state variables, which would require this contract be redeployed along with any of those other contracts, we can avoid that by adding a setter for the address so that the contract owner can update the address after launch.

- [ ] Add `ownerOnly` modified setter for the voucherKernel contract address var
- [ ] Add unit test for setter

bidyut-arianelabs:
This PR fixes https://github.com/bosonprotocol/contracts/issues/191

### DIFF
--- a/contracts/Cashier.sol
+++ b/contracts/Cashier.sol
@@ -38,6 +38,8 @@ contract Cashier is ICashier, UsingHelpers, ReentrancyGuard, Ownable, Pausable {
 
     event LogTokenContractSet(address _newTokenContract, address _triggeredBy);
 
+    event LogVoucherKernelSet(address _newVoucherKernel, address _triggeredBy);
+
     event LogWithdrawal(address _caller, address _payee, uint256 _payment);
 
     event LogAmountDistribution(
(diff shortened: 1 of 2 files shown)


cliffhall (approved):
LGTM! 👍

mischat (approved):
:shipit: :)

PR #234, opened 2021-10-26
Remove unnecessary require in VoucherKernel.createOrder

### LINKED ISSUE #202: Unnecessary Require statements

bidyut-arianelabs:
Fixes https://github.com/bosonprotocol/contracts/issues/202

### DIFF
--- a/contracts/VoucherKernel.sol
+++ b/contracts/VoucherKernel.sol
@@ -231,6 +231,7 @@ contract VoucherKernel is IVoucherKernel, Ownable, Pausable, ReentrancyGuard, Us
     nonReentrant
     onlyFromRouter
     returns (uint256) {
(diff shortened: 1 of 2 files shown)


cliffhall (approved):
LGTM! 👍

mischat (approved):
I am happy with this too. :)

zajck (approved):

PR #237, opened 2021-10-27
Removing unreachable code

zajck:
Removed 4 instances of unreachable code:
- VoucherKernel.generateTokenType was transformed to inline opearion
- require in Cashier._withdrawPayments() was removed
- two conditional blocks in voucherKernel.complain() were removed

I also updated the unit tests that test the complain period - mostly I just added the comment, so it's easier to understand what is the block timestamp.
In some instances I made the `advanceTimeSeconds` shorter, so the code is just barely after the point when period ends. Otherwise the test were not changed.


bidyut-arianelabs (approved):
Looks good to me 👍 

@hswopeams Can you please also have a look at the removal of code in `VoucherKernel.complain()`

hswopeams (approved):

PR #238, opened 2021-10-27
Revert when no state changes or events

zajck:
This PR adds require statements to functions, which until now gracefully executed even if they did emitted any event or change contract state.

Unit tests were adjusted so they expect revert reason instead of checking that state was not changed and no events were emitted.


cliffhall (dismissed):
LGTM! 👍

hswopeams (changes requested):
Request for one change and one question.

hswopeams (dismissed):

mischat (dismissed):
LGTM

bidyut-arianelabs (approved):
Looks good to me 👍

hswopeams (approved):

PR #239, opened 2021-10-27
Checking if receiver is a contract in method VoucherKernel.fillOrder

bidyut-arianelabs:
Refer: https://github.com/bosonprotocol/contracts/pull/223

```
checkOrderFillable and extract721 are always called only from one method (fillOrder) 
so we could move one of these checks there and get rid of the other.
```


zajck (changes requested):

zajck (approved):

cliffhall (approved):
LGTM! 👍

PR #243, opened 2021-10-27
Remediate DAI-02M

### LINKED ISSUE #242: DAI-02M: Rudimentary Sanitization of Cryptographic Components
## DAI-02M: Rudimentary Sanitization of Cryptographic Components

| [Input Sanitization](/appendix/finding-types#input-sanitization) | <span class="o-severity o-minor">Minor</span> | [DAITokenWrapper.sol:L64](https://github.com/bosonprotocol/contracts/blob/817b6814e4e93939fe237c525107f23ad67ce2da/contracts/DAITokenWrapper.sol#L64) |

### Description:

The sanitization applied to the `r` and `s` values of the `ecrecover` payload are rudimentary and do not actually cover the checks that should be applied based on the latest security guidelines.

### Example:

```solidity
/**
 * @notice Conforms to EIP-2612. Calls permit on token, which may or may not have a permit function that conforms to EIP-2612
 * @param owner Address of the token owner who is approving tokens to be transferred by spender
 * @param spender Address of the party who is transferring tokens on owner's behalf
 * @param value Number of tokens to be transferred
 * @param deadline Time after which this permission to transfer is no longer valid
 * @param v Part of the owner's signatue
 * @param r Part of the owner's signatue
 * @param s Part of the owner's signatue
 */
function permit(
    address owner,
    address spender,
    uint256 value,
    uint256 deadline,
    uint8 v,
    bytes32 r,
    bytes32 s
) 
    external
    override
    notZeroAddress(owner)
    notZeroAddress(spender)
{
    require(deadline == 0 || block.timestamp <= deadline, "PERMIT_EXPIRED");
    require(r != bytes32(0) && s != bytes32(0), "INVALID_SIGNATURE_COMPONENTS");
    uint nonce =  IDAI(daiTokenAddress).nonces(owner);
    IDAI(daiTokenAddress).permit(owner, spender, nonce, deadline, true, v, r, s);
    emit LogPermitCalledOnToken(daiTokenAddress, owner, spender, 0);    
}
```

### Recommendation:

We strongly recommend two sanitizations to be applied to the `v` and `s` members; the former should be validated to be equal to either `27` or `28` and the latter should be validated to exist in the lower half order of the ECDSA curve (i.e. less-than `0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A1`). Both checks are necessary to prevent repeat-able signatures from being accepted as valid.

## Remediation Strategy
After clarification from Omniscia, we will follow the auditor's recommendation.
- [x] Add require statements for both checks, similar to [the Open Zeppelin implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/04695aecbd4d17dddfd55de766d10e3805d6f42f/contracts/cryptography/ECDSA.sol#L49-L71).
- [x] Add unit tests for both revert reasons

cliffhall:
* In DAITokenWrapper.sol,
  - replace require that checks _r and _s with tests of _s and _v
  - as on the OZ contract reference mentioned in the comment, there is no need to test _r, as the DAI contract will do that implicitly.
  - this test is specifically to ensure non-replayability

* In revertReasons.ts,
  - replaced INVALID_SIGNATURE_COMPONENTS constant with INVALID_SIG_V and INVALID_SIG_S

* In 8_token_wrappers.ts,
  - replaced test 'Should revert if signatue portion r is invalid' with
  test 'Should revert if signature component s is invalid'
  - added test 'Should revert if signature component v is invalid'

* In 9_permit_with_token_wrapper.ts,
  - replaced test '[NEGATIVE] Should revert if token wrapper reverts because of invalid signature' with test '[NEGATIVE] Should revert if signature component s is invalid'
  - added test '[NEGATIVE] Should revert if signature component v is invalid'

* Update to package-lock.json,
  - had to run npm because hardhat-abi-exporter had been added, and apparently in the commit when it was added, package-lock.json was not updated

* This fixes #242


zajck (commented):
The code and tests looks good to me.

Only one thing - we have a convention that we do not check in `package-lock.json` unless also `package.json` is changed, so please remove that file.

zajck (approved):

bidyut-arianelabs (approved):
Looks good to me 👍

PR #222, opened 2021-10-22
Split 1155 721 contract

hswopeams:
This draft pull request includes the first steps for splitting the `ERC1155ERC721.sol` contract into two: `VoucherSets.sol `and `Vouchers.sol`.

To Do:

- Replace the local state vars and methods with inheriting from OZ in Vouchers and VoucherSets contracts
- Make sure any "special" (specific) functionality from `ERC1155ERC721.sol`
- Update unit tests
- Update deployment and verification scripts

The pipeline will fail because the unit tests haven't been updated!  This is a known issue.

PR #244, opened 2021-10-28
Check receiver whether it supports ERC-1155 / ERC-721

bidyut-arianelabs:
This PR adds better checks to see if the receiver supports ERC-721 / ERC-1155.

depends on https://github.com/bosonprotocol/contracts/pull/222

### DIFF
--- a/contracts/VoucherSets.sol
+++ b/contracts/VoucherSets.sol
@@ -174,6 +174,7 @@ contract VoucherSets is IVoucherSets, Ownable, ReentrancyGuard {
     /**
      * @notice Check successful transfer if recipient is a contract
      * @dev ERC-1155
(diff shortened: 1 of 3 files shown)


zajck (changes requested):

hswopeams (commented):
It looks like it matches the ERC1155ERC721 implementation. I have a question that is also relevant to the current ERC1155ERC721 implementation related to how the check functions are called. I noticed in the Vouchers.sol` _mint()` the call to `_doSafeTransferAcceptanceCheck()` is wrapped in a require. The same and similar calls in VoucherSets.sol are not. Is there a reason to use a require in some places and not others?

hswopeams (approved):

zajck (approved):

PR #246, opened 2021-10-28
Remove voucherKernel.burnSupplyOnPause

bidyut-arianelabs:
```
VoucherKernel includes the function `burnSupplyOnPause` which can only be called from the Cashier contract.
But Cashier never really calls it, so currently it is unusable.
```


cliffhall (dismissed):
Aside from the CI warnings about things defined but never used in unit tests, this looks good to me.

hswopeams (approved):

cliffhall (changes requested):
LGTM! 👍

cliffhall (approved):
LGTM! 👍 (looks like the previous approval didn't show up).

PR #247, opened 2021-10-28
This makes the build fail if our contracts get too big.

mischat:
Given that the size of our contracts have come up, I have added it to the build


zajck (approved):

bidyut-arianelabs (approved):
Looks good to me 👍

PR #251, opened 2021-10-29
This adds in the gas reporter to our ci

mischat:


zajck (dismissed):

hardhat.config.ts
```diff
@@ -73,8 +75,12 @@ const config: HardhatUserConfig = {
                 disambiguatePaths: false,
                 runOnCompile: false,
                 strict: true
+        },
+        gasReporter: {
```
hswopeams, line 79: This and the `contractSizer` config are indented too far

mischat: The indentation looks the same as the other config to me, perhaps I am missing something ? @hswopeams ? Can we accept this PR and then i can fix it in a subsequent one?

mischat: strange that the linter didn't pick it up... could it be tabs and spaces ....

mischat: Screenshot 2021-11-01 at 10 48 52

mischat: ok, so it is a tabs vs spaces thing. I have my IDE configured to represent tabs as 4 spaces and to never write out tabs, as tabs can be different on different machines. I will use tabs now, but we should fix this in the future and have our linter take care of this ...

mischat: Screenshot 2021-11-01 at 11 04 30

mischat: This should be fixed now @hswopeams

hswopeams (approved):

PR #252, opened 2021-11-01
Fix bug in timings

### LINKED ISSUE #201: _validFrom should be at least 5 minutes less than _validTo
https://github.com/bosonprotocol/contracts/blob/043d9304860ad0722eeef810d04a33418f9594c7/contracts/VoucherKernel.sol#L228

bidyut-arianelabs:
fixes https://github.com/bosonprotocol/contracts/issues/201


contracts/VoucherKernel.sol
```diff
@@ -237,6 +237,7 @@ contract VoucherKernel is IVoucherKernel, Ownable, Pausable, ReentrancyGuard, Us
         require(_validFrom <= _validTo, "INVALID_VALIDITY_FROM");
```
zajck, line 237: By adding require to line 240, this one becomes almost redundant -> when `_validTo >= _validFrom + 5 minutes` is true then  `_validFrom <= _validTo` is true as well.

I say almost because if you just remove line 237 there is an edge case, there `_validFrom` is set just below maxUint256 and adding 5 minutes would overflow, making require in 240 true even if `_validFrom` would actually be greater than `_validTo`. 

Suggestion: remove line 237 and add safemath to this line.

test/14_voucherKernel.ts
```diff
@@ -660,5 +660,71 @@ describe('VOUCHER KERNEL', () => {
         )
       ).to.be.revertedWith(revertReasons.INVALID_QUANTITY_LONG);
     });
+
+    it('[createTokenSupplyId] Should not revert if validFrom is 5 minutes or more than ValidTo', async () => {
+      await deployContracts();
+
+      // spoof boson router address
+      await contractVoucherKernel.setBosonRouterAddress(users.deployer.address);
+
```
zajck, line 669: Since 

      await deployContracts();
      // spoof boson router address
      await contractVoucherKernel.setBosonRouterAddress(users.deployer.address);
      
Repeates in last three tests, just put them togeether in describe and add that to beforeEach that does it.

zajck (changes requested):

test/14_voucherKernel.ts
bidyut-arianelabs: done

contracts/VoucherKernel.sol
bidyut-arianelabs: done

hswopeams (dismissed):

hswopeams:
Approved, assuming test failures get fixed

test/2_test_fullpath_with_permit.ts
```diff
@@ -502,7 +502,7 @@ describe('Cashier and VoucherKernel', () => {
             constants.PROMISE_DEPOSITBU1,
             constants.QTY_10
           )
-        ).to.be.revertedWith(revertReasons.INVALID_VALIDITY_FROM);
+        ).to.be.revertedWith(revertReasons.INVALID_VALIDITY_TO);
```
zajck, line 505: Revert reason does not match test description. But given that you made the tests for `VALID_FROM_MUST_BE_AT_LEAST_5_MINUTES_LESS_THAN_VALID_TO` (which have stricter requirements than this old test), this test here can be removed.

Same is true for the occurrences in the code below.

zajck (changes requested):

hswopeams (approved):

zajck (approved):

PR #256, opened 2021-11-01
Specify max tip on each tx

### LINKED ISSUE #255: Supply gas hints to transactions in deployment script
## Why are we considering fiddling with the gas during deployment?
Ropsten deployment failed last week and it was gas related. The automatic gas fee hints that ethers.js was providing did not work. This _might_ be because Ropsten often has erratic gas prices, we're not certain.

However if we provide the gas hints at deployment time, we can check the prevailing price and add enough wiggle room that they'll likely all go through.

cliffhall:
## This PR has been replaced with PR #262 
- Updated .env.example
  - added MAX_TIP

- In deploy.ts,
  - in DeploymentExecutor
    - added maxTip and txOptions
    - in constructor
      - set maxTip to process.env.MAX_TIP
      - set txOptions to an object mapping maxPriorityFeePerGas to this.maxTip
      - in all deploy and method calls, pass this.txOptions as last or sole argument

- This fixes #255


cliffhall:
@hswopeams have a look at #255 for the rundown on this PR. Also, give it a try. I couldn't deploy to hardhat network because of an unrelated type thing in the deployment script that was present on main when I branched.

mischat:
is it a linting issue that is making this fail ?

mischat:
> is it a linting issue that is making this fail ?

I have fixed this by running `npm run tests:format-fix` hopefully this does the trick

hswopeams (changes requested):
This looks good, and I can run it locally against a forked node.  One additon: The txOptions object needs to be added to the transactions in the general `setDefaults` function that starts on line 65. That one gets called for all environments. The  `setDefaults` functions in the Prod and NonProd Executors set environment-specific settings.

cliffhall:
> The txOptions object needs to be added to the transactions in the general `setDefaults` function that starts on line 65. 

@hswopeams just updated these, good catch.

cliffhall:
Head on over to PR #262

cliffhall:
Head on over to PR #262

PR #257, opened 2021-11-02
Setting contracts addresses in constructors

zajck:
In this PR:

- updated VoucherKernel, Cashier, Gate, VoucherSets and Voucher. All corresponding contracts are now set in the constructor and do not have to be set right after the deployment
- updated unit test to match. Some were removed, since are not applicable anymore (contracts are always set)
- updated deploy and verify script


hswopeams:
This is a general comment:  the `Cashier.sol` has a `setVoucherKernelAddress` function, but it is the only set function that is not in the `ICahier.sol`. To keep everything consistent, it should be added.

hswopeams (commented):
See general comment above

hswopeams (approved):

bidyut-arianelabs (approved):
Looks good to me 👍

PR #259, opened 2021-11-02
Deployment setter parameters in env

zajck:
In this PR:
- `.env` has additional variables for setting ETH and token limits and for wait periods.
- separated token limits for dai and boson token
- added setETHlimit in setters


cliffhall (approved):
Looks good to me but will definitely have some conflicts with #262

hswopeams (approved):

PR #260, opened 2021-11-02
Adding whenPaused modifier to contract address setters

### LINKED ISSUE #194: TRY-01M: Overly Centralized Wrapper Control
## TRY-01M: Overly Centralized Wrapper Control

| [Logical Fault](/appendix/finding-types#logical-fault) | <span class="o-severity o-minor">Minor</span> | [TokenRegistry.sol:L83-L91](https://github.com/bosonprotocol/contracts/blob/817b6814e4e93939fe237c525107f23ad67ce2da/contracts/TokenRegistry.sol#L83-L91) |

### Description:

The `setTokenWrapperAddress` function is meant to be invoked to maintain the list of actively supported tokens in the Boson Protocol and allow custom wrappers to be utilized for non-EIP-2612 conformant tokens, however, in doing so the `owner` has the power to change the wrapper at will.

### Example:

```solidity
function setTokenWrapperAddress(address _tokenAddress, address _wrapperAddress) 
    external
    override
    onlyOwner
    notZeroAddress(_tokenAddress)
{
    tokenWrappers[_tokenAddress] = _wrapperAddress;
    emit LogTokenWrapperChanged(_wrapperAddress, owner());
}
```

### Recommendation:

We advise an additional `require` check to be introduced in the function that ensures the wrapper hasn't been previously set to prevent them from being set when a voucher transaction is in the mempool or after a voucher order has been created.

## Remediation Strategy
We will use a multisig wallet for `onlyOwner` modified methods. However the caution about changing this address when a transaction is in the mempool is important to consider.

We could and probably _should_ make **this** and all of the `onlyOwner` modified address setters in the system also have the `whenPaused` modifier. This would minimize the possibility of transactions in the mempool hitting a contract configuration that has changed since it was added.

- [ ] Add the `whenPaused` modifier to `setTokenWrapperAddress`
- [ ] Find and make the same change to any other `onlyOwner` modified address setters in the system, keeping in mind that some of those setters may not be implemented yet as there are a few pending issues ([178](https://github.com/bosonprotocol/contracts/issues/178), [182](https://github.com/bosonprotocol/contracts/issues/182), [191](https://github.com/bosonprotocol/contracts/issues/191)) which will add such setters.

zajck:
`whenPaused` is added to all setters that are setting collaborators' addresses.

`BosonRouter.pause()` and `BosonRouter.pause()` were modified to prevent deadlock that would happen after BosonRouter address is changed (new contract would be unable to unpause since it is not paused, while old contract cannot unpause `cashier` and `voucherKernel` since it is not authorized anymore)


test/1_test_fullpath.ts
```diff
@@ -1333,16 +1337,24 @@ describe('Voucher tests', () => {
...
       await contractCashier.setBosonRouterAddress(
         contractMockBosonRouter.address
       );
 
+      await contractVoucherKernel.setBosonRouterAddress(
+        contractMockBosonRouter.address
+      );
+
+      await contractMockBosonRouter.pause();
+      await contractMockBosonRouter.unpause();
```
cliffhall, line 1351: Why does this test pause then immediately unpause the mock boson router without doing anything inbetween? Add a comment if this isn't an oopsy.

test/1_test_fullpath.ts
```diff
@@ -626,13 +626,17 @@ describe('Voucher tests', () => {
...
+
       await contractVoucherKernel.setBosonRouterAddress(
         contractMockBosonRouter.address
       );
 
       await contractCashier.setBosonRouterAddress(
         contractMockBosonRouter.address
       );
+      await contractMockBosonRouter.pause();
+      await contractMockBosonRouter.unpause();
```
cliffhall, line 639: Why does this test pause then immediately unpause the mock boson router without doing anything inbetween? Add a comment if this isn't an oopsy.

test/2_test_fullpath_with_permit.ts
```diff
@@ -2216,13 +2216,16 @@ describe('Cashier and VoucherKernel', () => {
...
+        await contractBosonRouter.pause();
         await contractVoucherKernel.setBosonRouterAddress(
           contractMockBosonRouter.address
         );
 
         await contractCashier.setBosonRouterAddress(
           contractMockBosonRouter.address
         );
+        await contractMockBosonRouter.pause();
+        await contractMockBosonRouter.unpause();
```
cliffhall, line 2228: Why does this test pause then immediately unpause the mock boson router without doing anything inbetween? Add a comment if this isn't an oopsy.

test/7_admin_functionality.ts
```diff
@@ -180,7 +180,7 @@ describe('Admin functionality', async () => {
   }
 
   describe('Cashier', () => {
-    before(async () => {
+    beforeEach(async () => {
```
cliffhall, line 183: Good catch. 👍

cliffhall (changes requested):
Mostly good. Just a few spots in tests that need a comment or removal of a line.

test/1_test_fullpath.ts
zajck: It's needed in the current implementation. I added a comment explaining why.

test/1_test_fullpath.ts
zajck: It's needed in the current implementation. I added a comment explaining why.

test/2_test_fullpath_with_permit.ts
zajck: It's needed in the current implementation. I added a comment explaining why.

cliffhall (dismissed):
LGTM! 👍

hswopeams (changes requested):
`Cahsier.setVoucherSetTokenAddress` is missing the `whenPaused` modifier

hswopeams:
I noticed that VoucherKernel does not have setVoucherTokenAddress or setVoucherSetTokenAddress functions, even though these addresses are present in the contract. It is the only contract that has address variables that are not settable. It seems like these should be added with modifier whenPaused.

zajck:
Beside adding setters in `VoucherKernel` I extneded `Vouchers` and `VoucherSets` to be `Pausable` so I could add whenPaused modifiers there too

zajck:
Closing since CI is producing different results that local test run.
Let's see if new PR resolves the issue.

PR #261, opened 2021-11-02
Contract uri for open sea collections

### LINKED ISSUE #228: Add OpenSea contract-level metadata support
The OpenSea collections page for a contract is most easily controlled by adding a contractURI method to the token contract which returns a URI pointing to an off-chain metadata file about the collection.

The method looks like this:

```solidity
function contractURI() public view returns (string memory) {
        return "https://bosonprotocol.io/contractMetadata.json";
    }
}
```

The metadata looks like this:
```JSON
{
  "name": "Boson Protocol Vouchers",
  "description": "Vouchers represent the right to purchase.... PLEASE ADD GOOD DESC HERE",
  "image": "https://bosonprotocol.io/voucherImage.png",
  "external_link": "https://bosonprotocol.io",
  "seller_fee_basis_points": 0, # Indicates a 1% seller fee.
  "fee_recipient": "" # Where seller fees will be paid to.
}
```

- [ ] Add `contractURI` method to Vouchers.sol (Currently ERC1155721.sol in main branch but soon to be split)
- [ ] Compose metadata and make sure whatever team is over serving the URL or uploading the metadata to IPFS has the baton

bidyut-arianelabs:
fixes https://github.com/bosonprotocol/contracts/issues/228


scripts/deploy.ts
```diff
@@ -84,6 +84,16 @@ class DeploymentExecutor {
       event.args._approved
     );
 
+    tx = await this.vouchers.setContractUri(process.env.VOUCHERS_CONTRACT_URI);
```
cliffhall, line 87: After merging #262, add `this.txOptions as final argument to this function, like so:

```javascript
tx = await this.vouchers.setContractUri(process.env.VOUCHERS_CONTRACT_URI, this.txOptions);
```

cliffhall (changes requested):
Looks good, but I'd like to have PR #262 merged first, and then there's the one request below.

mischat:
This works for me, a part of me thought we could just set it in the constructor but this looks fine. I guess we just need to do the same work on the vouchersets and we are good.

scripts/deploy.ts
mischat: The above PR is now merged, and tested on rinkeby, i guess we are good to go on his side of things now.

bidyut-arianelabs: I have changed the `scripts/deploy.ts` where `this.txOptions` isn't available.
`this.txOptions` is for the `deploy.ts` file where we do not deploy `Vouchers` and `VoucherSets`.

zajck: `deploy.ts` in root was created in https://github.com/bosonprotocol/contracts/pull/262 but that file is not something that `npx hardhat deploy` (or `npm run contracts:migrate`) actually use. That `deploy.ts` is also obsolete (has still `erc1155erc721` contract and does not set contract addresse via constructors).

Actually this `scripts/deploy.ts` should have `this.txOptions` since this is the script used in deployment.

bidyut-arianelabs: Done in https://github.com/bosonprotocol/contracts/pull/261/commits/b93e54d0836fb56d447bc93d3a5565fe58f0e22d

zajck (dismissed):

mischat (dismissed):
This looks good to me!

scripts/deploy.ts
mischat: @zajck can you resolve this conversation, given timezones with cliff and given you commented on this ?

hswopeams: deploy.ts should not be in the root directory

mischat: @bidyut-arianelabs ^^

bidyut-arianelabs: Removed in https://github.com/bosonprotocol/contracts/pull/261/commits/facd919b6636b44e5c29dc1b80708d3b3fe77a99

zajck (approved):

mischat (approved):
LGTM

mischat:
@cliffhall can you unblock this, as we would like to get this into main, so that we can create a new RC release

cliffhall:
> @cliffhall can you unblock this, as we would like to get this into main, so that we can create a new RC release
@mischat Not sure how I'm blocking.

hswopeams:
> > @cliffhall can you unblock this, as we would like to get this into main, so that we can create a new RC release
> > @mischat Not sure how I'm blocking.

@cliffhall You requested changes, so it can't be merged until you approve it.

cliffhall (approved):
LGTM! 👍

PR #262, opened 2021-11-02
Replacement PR: Specify max tip on each tx 

### LINKED ISSUE #255: Supply gas hints to transactions in deployment script

cliffhall:
## Replacement PR. Supersedes #256  

- In .env.example
  - Changed comment identifiers to # from // (this parsed by shell)
  - added `MAX_TIP`

- In deploy.ts,
  - in DeploymentExecutor
    - added `maxTip` and `txOptions`
    - in constructor
      - set maxTip to process.env.MAX_TIP
      - set txOptions to an object mapping `maxPriorityFeePerGas `to `this.maxTip`
    - in all deploy and method calls, pass `this.txOptions` as last or sole argument
        
- This fixes #255


cliffhall:
This is a replacement for PR #256. I bungled the branch merging from main (there were conflicts so I tried to revert the merge and defer reconciliation but it didn't go so well) and and since there were only two files, I just made a new branch and replacement PR.

.env.example
```diff
@@ -1,16 +1,23 @@
...
-//DAI token Rinkeby address
+# DAI token Rinkeby address
 DAI_TOKEN=0x6A9865aDE2B6207dAAC49f8bCba9705dEB0B0e6D
 
-//metadata URIs
+# metadata URIs
 CONDITIONAL_COMMIT_TOKEN_METADATA_URI=changeme
-VOUCHERSETS_METADATA_URI=changeme
-VOUCHERS_METADATA_URI=changeme
+METADATA_BASE=changeme
```
hswopeams, line 18: These still aren't correct. It should be 
OUCHERSETS_METADATA_URI=changeme
VOUCHERS_METADATA_URI=changme

METADATA_BASE should be removed

hswopeams (changes requested):
See comment about .env.example

.env.example
cliffhall: Done.

hswopeams (dismissed):

hswopeams (approved):

hswopeams:
Approved, assuming tests will pass.

mischat (approved):
Let's do this.

PR #266, opened 2021-11-02
This updates the number of times the optimizer runs, this is the default

mischat:
value used.


zajck (approved):

bidyut-arianelabs (approved):
Looks good to me 👍

hswopeams:
Merging because it has two approvals

PR #267, opened 2021-11-03
Some clarity to the .env.example file

mischat:


.env.example
```diff
@@ -13,19 +13,19 @@ BOSON_TOKEN=0x056b1ce5510ceefa6c33600551e55333aff8296f
 # DAI token Rinkeby address
```
zajck, line 13: This should be without rinkeby.

Also line 10 `# local hardhat address #9` is not ok description.

zajck (changes requested):

.env.example
mischat: @zajck not sure what you mean, re: rinkeby ? what DAI token address is that ? which chain ?

And i can update line 10

zajck: I've discussed that with @hswopeams 

The values are ok, it just wasn't clear for me, if the defaults have any meaning. And they might confuse someone when they see them first.

So I would just add comment like "Default values for BOSON_TOKEN and DAI_TOKEN are required for local deployment. To do deployment on chain, adjust them accordingly" (or something similar that tells the same)

mischat: @zajck updated.

zajck (dismissed):

.env.example
```diff
@@ -7,25 +7,24 @@ CMC_API_KEY=000000000000000000000000000000000000000
...
-VOUCHERSETS_METADATA_URI=changeme
-VOUCHERS_METADATA_URI=changme
+CONDITIONAL_COMMIT_TOKEN_METADATA_URI=https://metadata.bosonportal.example/quest/{id}.json
+VOUCHERSETS_METADATA_URI=https://metadata.boson.example/voucherset/{id}.json
+VOUCHERS_METADATA_URI=https://metadata.boson.example/voucher/
 
 # Max tip in gwei - use "Fastest" setting from: https://www.etherchain.org/api/gasPriceOracle
-MAX_TIP = 13.1
\ No newline at end of file
+MAX_TIP = 13.1# API Keys
```
hswopeams, line 30: why is # API Keys needed?

hswopeams (changes requested):

.env.example
mischat: :)

mischat: fixed ...

zajck (approved):

hswopeams (approved):

PR #272, opened 2021-11-04
Fixes npm publish workflow

dohaki:
- Add env vars with fake deployers PKs to npm publish workflows
- Add deployed contract addresses to package


mischat (approved):
This looks good, the issues with the build was that missing ENV variables, so my guess is that this fixes it. thanks!

mischat:
Having spoken to @dohaki I am going to merge this, cause if i don't we get 1.0.0 pushed to npm.

PR #274, opened 2021-11-08
Run code coverage when PR merged into main branch

bidyut-arianelabs:
This PR makes it so that the code coverage gets run when PR is merged into `main` halving the time it will take to run the build on submitted PRs.

PR #275, opened 2021-11-09
Uploading the code coverage report and Gas usage report

bidyut-arianelabs:
Depends on https://github.com/bosonprotocol/contracts/pull/274

## 1. Code coverage report uploaded.
---
Screenshot 2021-11-09_18-04-08-571

### Once downloaded, the Code coverage report file can be viewed in the browser like this.
Screenshot 2021-11-09_18-11-38-581

---
## 2. Gas report uploaded (Generated whenever the unit test is executed).
---
Screenshot 2021-11-09_18-03-43-936

### Once downloaded, the Gas reporter file can be viewed in the terminal using the `cat` command like this.
Screenshot 2021-11-09_18-05-54-048

---

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,6 +36,7 @@ jobs:
     env:
       PROTOCOL_DEPLOYER_PRIVATE_KEY: 123456789abcdef123456789abcdef123456789abcdef123456789abcdef1234
       CC_TOKEN_DEPLOYER_PRIVATE_KEY: 123456789abcdef123456789abcdef123456789abcdef123456789abcdef1234
+      CMC_API_KEY: ${{ secrets.CMC_API_KEY }}
     name: test
     steps:
       - name: Checkout repository
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,7 +13,7 @@ jobs:
     env:
       PROTOCOL_DEPLOYER_PRIVATE_KEY: 123456789abcdef123456789abcdef123456789abcdef123456789abcdef1234
       CC_TOKEN_DEPLOYER_PRIVATE_KEY: 123456789abcdef123456789abcdef123456789abcdef123456789abcdef1234
-      CMC_API_KEY: ${{ secrets.CMC_API_KEY_NEW }}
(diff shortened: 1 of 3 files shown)


cliffhall (changes requested):
You see a dashes in the rightmost column of the gas report because it doesn't have a fiat ETH price to extend gas usage. 

I would like to see the unit test / gas reporting process use a coinmarketcap key so that the current price of ETH is shown and the cost of each method call and deployment is extended to a fiat price.

hswopeams (dismissed):

NFTanaconda:
nice

cliffhall (dismissed):

bidyut-arianelabs:
PR is now ready for review.
Data from coinmarketcap is now visible. refer https://github.com/bosonprotocol/contracts/actions/runs/1654441004

hswopeams (dismissed):

.github/workflows/ci.yaml
```diff
@@ -13,6 +13,7 @@ jobs:
     env:
       PROTOCOL_DEPLOYER_PRIVATE_KEY: 123456789abcdef123456789abcdef123456789abcdef123456789abcdef1234
       CC_TOKEN_DEPLOYER_PRIVATE_KEY: 123456789abcdef123456789abcdef123456789abcdef123456789abcdef1234
+      CMC_API_KEY: ${{ secrets.CMC_API_KEY_NEW }}
```
cliffhall, line 16: So we determined that there was some clobbering or mismatch of the key in the secrets object. But can we change the variable to be something like COINMARKETCAP_API_KEY? The names above it are verbose, so it doesn't need to be abbreviated or have NEW in it.

cliffhall (changes requested):

.github/workflows/ci.yaml
bidyut-arianelabs: I agree. I have renamed the secret variable to `COINMARKETCAP_API_KEY`. PR is now ready for review. @cliffhall

hswopeams (approved):

cliffhall (approved):
LGTM! 👍

.github/workflows/ci.yaml
cliffhall: Cool, thanks

PR #281, opened 2021-11-19
Deploy mock token locally

hswopeams:
This PR changes the deployment script so that a mock Boson Token is deployed when running locally. The README was also changed to reflect this. In addition, the SIXTY_SECONDS constant was removed because it is no longer used.

### DIFF
--- a/scripts/deploy.ts
+++ b/scripts/deploy.ts
@@ -311,6 +311,24 @@ class DeploymentExecutor {
     }
   }
 
+  async deployMockToken() {
+    //only deploy the mock for local environment
+    if(this.env == 'hardhat') {
+      console.log("Deploying mock Boson Token");
+
+      const MockBosonToken = await ethers.getContractFactory('MockERC20Permit');
+      const mockBosonToken = await MockBosonToken.deploy(
+        "Mock Boson Token",
+        "BOSON"
+      );
+
+      await mockBosonToken.deployed();
+      this.boson_token = mockBosonToken.address;
+    } else {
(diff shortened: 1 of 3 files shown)


scripts/deploy.ts
```diff
@@ -311,6 +311,24 @@ class DeploymentExecutor {
     }
   }
 
+  async deployMockToken() {
+    //only deploy the mock for local environment using default deployer address
+    if (this.env == 'hardhat') {
```
zajck, line 316: `if` is not really needed here, since the same check is done before calling `deployMockToken`.
Anyway, it does not hurt if you keep it as is (my approval will not depend on this).

zajck (dismissed):
It looks ok to me.

Just one thing to consider. If developer uses `npx hardhat deploy` instead of npm scripts, there is a chance that they forget to add `env` when deploying to testnets, then `env` defaults to `hardhat`, therefore deploying mock boson contract to testnet. I don't think it's critical, yet it does not seem a desired behaviour. I don't think the changes are needed (since this "problem" is a result of developer mistake), but if you wish, you can address it (anyway, I'm approving it even if you don't)

bidyut-arianelabs (dismissed):
Looks good to me 👍

scripts/deploy.ts
hswopeams: I modified the script to check network instead of env. I took the` if` out of the main function and left it in here.

zajck (approved):

bidyut-arianelabs (approved):
Looks good to me 👍

PR #282, opened 2021-11-24
Omni token gate

cliffhall:
Add support for standard ERC721, ERC20, and ERC1155 tokens as conditional token for gate check, while not breaking anything with existing implementation.

- `Gate` refactored 
  -  Events declared in the contract rather than interface
  - term is **conditional token** rather than **non-transferable token** 
  - `TokenType` is an enum TOKEN or MULTI_TOKEN (ERC20/721 = TOKEN, ERC115.= MULTI_TOKEN)
  - extra constructor arg called `_conditionalTokenType` (no setter/getter yet)
  - `setConditionalContract` setter takes contract address and type
  -  `getConditionalContract` getter returns contract address and type
  - interfaces for Token and MultiToken types that have the proper `balanceOf` signature
  - check function now calls `balanceOf` appropriately based on `conditionalTokenType`
 
- `IGate` refactored
  - Events moved to contract
  - Unnecessary function signatures removed, i.e., everything not called by the other contracts

- Gate unit tests still work
- Conditional Commit unit tests work with all conditional token types 
    - ERC20
    - ERC721
    - ERC1155
    - ERC1155NonTransferrable (Quest)


zajck (commented):
Looks good to me :+1: 

In BosonRouter.sol there are still a few places with reference to `ERC115NonTransferrable` instead of `conditional`
Lines 213, 298, 402, 487

cliffhall:
Thanks @zajck. I was focuse on the gate and didn't notice (or expect) references to the nontransferable there. Will clean up this morning.

NFTanaconda:
cool

test/12_conditional_commit.ts
```diff
@@ -38,379 +63,507 @@ let BosonRouter_Factory: ContractFactory;
...
+                    [
+                      constants.PROMISE_VALID_FROM,
+                      constants.PROMISE_VALID_TO,
+                      constants.PROMISE_PRICE1,
+                      constants.PROMISE_DEPOSITSE1,
+                      constants.PROMISE_DEPOSITBU1,
+                      constants.QTY_10,
+                    ],
+                    contractGate.address,
+                    constants.NFT_TOKEN_ID,
```
zajck, line 531: It would be nice to rename `NFT_TOKEN_ID` to `CONDITIONAL_TOKEN_ID` to be in line with other changes (applies in other places as well)

test/12_conditional_commit.ts
```diff
@@ -38,379 +63,507 @@ let BosonRouter_Factory: ContractFactory;
...
+      await contractGate.unpause();
+
+      // Mint an ERC1155NonTransferable token for buyer
+      await contractERC1155NonTransferable.mint(
+        users.buyer.address,
+        constants.NFT_TOKEN_ID,
+        constants.ONE,
+        constants.ZERO_BYTES
+      );
+      //console.log("Setup Quest");
```
zajck, line 253: comment can be removed

test/12_conditional_commit.ts
```diff
@@ -38,379 +63,507 @@ let BosonRouter_Factory: ContractFactory;
...
+      await contractGate.pause();
+      await contractGate.setConditionalTokenContract(
+        contractSampleERC20.address,
+        constants.TOKEN_TYPE.TOKEN
+      );
+      await contractGate.unpause();
+
+      // Mint an ERC20 token for buyer
+      await contractSampleERC20.mint(users.buyer.address, constants.ONE);
+      //console.log("Setup ERC20");
```
zajck, line 267: comment can be removed

test/12_conditional_commit.ts
```diff
@@ -38,379 +63,507 @@ let BosonRouter_Factory: ContractFactory;
...
+      );
+      await contractGate.unpause();
+
+      // Mint an ERC1155 token for buyer
+      await contractSampleERC1155.mint(
+        users.buyer.address,
+        constants.NFT_TOKEN_ID,
+        constants.ONE
+      );
+      //console.log("Setup ERC1155");
```
zajck, line 299: comment can be removed

test/12_conditional_commit.ts
```diff
@@ -38,379 +63,507 @@ let BosonRouter_Factory: ContractFactory;
...
+      await contractGate.pause();
+      await contractGate.setConditionalTokenContract(
+        contractSampleERC721.address,
+        constants.TOKEN_TYPE.TOKEN
+      );
+      await contractGate.unpause();
+
+      // Mint an ERC721 token for buyer
+      await contractSampleERC721.mint(users.buyer.address);
+      //console.log("Setup ERC721");
```
zajck, line 281: comment can be removed

zajck (changes requested):
Very elegant solution for all different token types.
But if I'm not mistaken, some of the tests inside a loop does not really depend on the token type, therefore the same tests are performed 4 times instead of just once. This happens for all cases when testing the creation of vouchers (`TOKEN SUPPLY CREATION WITH TOKEN CONDITIONAL COMMIT (Create Voucher Set)`).

Second loop inside `VOUCHER CREATION (Commit to buy)` seems ok.

test/12_conditional_commit.ts
hswopeams: Same in the `it `

test/12_conditional_commit.ts
```diff
@@ -447,280 +602,295 @@ describe('Create Voucher sets and commit to vouchers with token conditional comm
...
+            )[0];
+
+            assert.isTrue(
+              sellerVoucherSetsBalance.eq(constants.QTY_10),
+              'VoucherSets seller balance mismatch'
+            );
+          });
+
+          describe('Flow with automatic gate.registerVoucherSetId', () => {
+            it('Should be able to create Voucher with gate address and non empty nft token id', async () => {
```
hswopeams, line 820: Change to conditional token Id

test/12_conditional_commit.ts
```diff
@@ -772,306 +944,301 @@ describe('Create Voucher sets and commit to vouchers with token conditional comm
...
```
hswopeams, line 1170: Change to conditional token Id

test/12_conditional_commit.ts
```diff
@@ -1121,270 +1290,251 @@ describe('Create Voucher sets and commit to vouchers with token conditional comm
...
```
hswopeams, line 1483: Change to conditional token Id

hswopeams (changes requested):
Looks pretty good. I have a few questions/comments/requests for change.

cliffhall:
> But if I'm not mistaken, some of the tests inside a loop does not really depend on the token type, therefore the same tests are performed 4 times instead of just once. This happens for all cases when testing the creation of vouchers (`TOKEN SUPPLY CREATION WITH TOKEN CONDITIONAL COMMIT (Create Voucher Set)`).
> 
> Second loop inside `VOUCHER CREATION (Commit to buy)` seems ok.

@zajck  Yes, this was my expectation as well, but I wasn't 100% certain if there was anything in those voucher set creation tests that would possibly break. So erring on the side of caution, I put the loop on those tests as well. I will take them out.

test/12_conditional_commit.ts
cliffhall: Good call. I replaced `NFT_TOKEN_ID` with `CONDITIONAL_TOKEN_ID` throughout the codebase which included `constants.ts`, `10_erc155_non_transferrable.ts`, `11_gate.ts`, and `12_conditional_commit.ts`

test/12_conditional_commit.ts
cliffhall: Had one of these for all the token types. Removed them all

contracts/Gate.sol
```diff
@@ -2,35 +2,81 @@
...
+        uint256 indexed _tokenIdSupply,
+        uint256 indexed _conditionalTokenId
+    );
+
+    event LogUserVoucherDeactivated(
+        address indexed _user,
+        uint256 indexed _tokenIdSupply
+    );
+
     mapping(uint256 => uint256) private voucherToToken;
```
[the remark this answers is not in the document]
cliffhall: Another good catch. Yes it was called `voucherToToken` before, but is using the voucher set's token id for the mapping to the conditional token. Changing to `voucherSetToToken`.

contracts/mocks/SampleERC1155.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.7.6;
+
+import "@openzeppelin/contracts/token/ERC1155/ERC1155.sol";
+
+/**
+ * @title Sample ERC1155 NFT for Unit Testing
+ * @author Cliff Hall
+ */
+contract SampleERC1155 is ERC1155 {
```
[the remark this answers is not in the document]
cliffhall: Yes. This was copypasta from another project, should have aligned the naming, thanks.

test/12_conditional_commit.ts
cliffhall: Replaced "nft token id" with "conditional token id" in several locations throughout the contract.

test/11_gate.ts
```diff
@@ -199,8 +230,9 @@ describe('Gate contract', async () => {
     );
 
     await gate.pause();
-    await gate.setNonTransferableTokenContract(
-      contractERC1155NonTransferable.address
+    await gate.setConditionalTokenContract(
+      contractERC1155NonTransferable.address,
```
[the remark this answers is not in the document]
cliffhall: There is actually only one test that needs to be repeated over all the token types and that's the one that calls the `Gate.check()` method. Elsewhere it checks that you can't set a zero address or an invalid token type. 

I've set up a separate `describe` block and refactor/moved the check test into it, wrapped by a loop over the token types similar to what is implemented in the conditional commit tests.

contracts/BosonRouter.sol
```diff
@@ -1192,18 +1192,18 @@ contract BosonRouter is
...
      * ownership of which is a condition for committing to redeem a voucher in the voucher set created by this function.
-     * @param _nftTokenId Id of the NFT (ERC115NonTransferrable) token, ownership of which is a condition for committing to redeem a voucher
+     * @param _conditionalTokenId Id of the conditional token, ownership of which is a condition for committing to redeem a voucher
      * in the voucher set created by this function.
      */
-    function finalizeConditionalOrder(uint256 _tokenIdSupply, address _gateAddress, uint256 _nftTokenId) internal {
+    function finalizeConditionalOrder(uint256 _tokenIdSupply, address _gateAddress, uint256 _conditionalTokenId) internal {
         voucherSetToGateContract[_tokenIdSupply] = _gateAddress;
 
         emit LogConditionalOrderCreated(_tokenIdSupply, _gateAddress);
```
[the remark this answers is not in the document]
cliffhall: Agreed. But it won't always be present.

- For ERC20, the contract is the token, so it would never have a tokenId registered
- For ERC1155, it needs a tokenId always.
- For ERC721, it's interesting because you could target
  - the owner of a **specific** NFT (Avastar number 153) 
  - holders of **any** NFT on the conditional token contract
  
 I am only testing the use of a conditional token ID on the Quest and ERC1155.

test/11_gate.ts
```diff
@@ -72,23 +82,44 @@ describe('Gate contract', async () => {
     contractBosonRouter: BosonRouter,
     contractTokenRegistry: TokenRegistry,
     contractBSNTokenPrice: MockERC20Permit,
-    contractBSNTokenDeposit: MockERC20Permit;
+    contractBSNTokenDeposit: MockERC20Permit,
+    contractSampleERC20: SampleERC20,
+    contractSampleERC721: SampleERC721,
+    contractSampleERC1155: SampleERC1155;
```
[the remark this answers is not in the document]
cliffhall: I added a cycle of tests over all token types that calls the `Gate.check()` method.

zajck (changes requested):
Beside the comments in code I think it would also be good at this point to update the definition of  `Gate` in `docs/contracts/overview.md`. And maybe also all other paragraphs in this document where it describes conditional commit.

contracts/BosonRouter.sol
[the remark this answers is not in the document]
hswopeams: The code change looks fine, but it looks like the test cases that check LOG_CONDITIONAL_ORDER_CREATED in 12_conditional_commit.ts weren't changed to check for the conditional token Id.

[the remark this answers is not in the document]
hswopeams: The code changes were made, but it looks like the LOG_CONDITIONAL_ORDER_CREATED assertions in 12_conditional_commit.ts weren't updated to check the conditional token Id

hswopeams (changes requested):
Still one outstanding change that needs to include the unit test code. See comments.

hswopeams (approved):
Reviewed changes @zajck made according to his requests for changes so that we can get this PR approved

zajck (approved):

mischat:
am going to look at this over the weekend, awesome stuff!

PR #289, opened 2021-12-23
Adapt the protocol to support meta-transactions

levalleux-ludo:
No expectation to be merged as is.
Just a draft to show and demonstrate how to adapt the protocol contracts to support meta-transactions, with the protocol contracts deployed on a layer2 chain.
Limited to the methods of the BosonRouter contract (assuming that if a user needs to call methods of other contracts, it's still possible to switch the wallet to the layer2 to call them directly)
Support of tokens without Permit method have been added too. Methods requestCreateOrderTKNTKNNoPermit, requestCreateOrderTKNTKNNoPermitConditional, and requestVoucherTKNTKNNoPermit have been created to clarify the calls (even if, basically, it would be enough to call existing methods with null values for deadline and the signature fields).
Support of the tokens without Permit is possible thanks to a specific token Wrapper contract (NoPermitTokenWrapper), assuming that the approve method has been called beforehand to allow the BosonRouter contract to transfer user"s funds.

### DIFF
--- a/contracts/BosonRouter.sol
+++ b/contracts/BosonRouter.sol
@@ -15,6 +15,8 @@ import "./interfaces/IGate.sol";
 import "./interfaces/ITokenWrapper.sol";
 import {PaymentMethod} from "./UsingHelpers.sol";
 import "./libs/SafeERC20WithPermit.sol";
+import "./MetaTransactionEIP712Receiver.sol";
+
 
 /**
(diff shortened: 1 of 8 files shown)


hswopeams:
@levalleux-ludo Do we need the `ChildDAITokenWrapper` purely because the DAI token has a different address for it's child token on Polygon?

cliffhall:
We decided to forgo operations on Polygon for v1.

PR #292, opened 2022-01-18
Modify call to permit to not require token wrapper

hswopeams:
This enables tokens to be supported without first having to be registered with a wrapper if they conform to EIP-2612. Any token that has a permit function that does NOT conform to EIP-2612 (such as DAI) must have a registered wrapper

### DIFF
--- a/contracts/BosonRouter.sol
+++ b/contracts/BosonRouter.sol
@@ -883,9 +883,12 @@ contract BosonRouter is
     ) internal {
         address tokenWrapper = ITokenRegistry(tokenRegistry)
             .getTokenWrapperAddress(_token);
-      
-        if(tokenWrapper != address(0)) { //tokens like DAI don't conform to EIP-2612, so a token wrapper is needed
-            ITokenWrapper(tokenWrapper).permit(
+
+        bool isWrapped = ( tokenWrapper != address(0) );
+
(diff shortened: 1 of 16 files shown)


zajck (dismissed):
Looks good to me.

contracts/BosonRouter.sol
```diff
@@ -883,19 +883,29 @@ contract BosonRouter is
...
-            _tokenOwner,
-            _spender,
-            _value,
-            _deadline,
-            _v,
-            _r,
-            _s
-        );
+      
+        if(tokenWrapper != address(0)) { //tokens like DAI don't conform to EIP-2612, so a token wrapper is needed
```
cliffhall, line 887: Since the method on the `ITokenWrapper.permit` and the `IERC20WIthPermit.permit` have the same signature and are interchangeable for the purpose of this call, all we need to do here is alternate based on the expression. This if/else block could be replaced by:

```solidity
bool isWrapped = ( tokenWrapper != address(0) );
IERC20WithPermit( isWrapped ? tokenWrapper : _token );
```

cliffhall (changes requested):
Just the one suggestion in BosonRouter.sol.

contracts/BosonRouter.sol
hswopeams: Modified.

cliffhall:
We chose not to implement this feature for v1.

PR #295, opened 2022-01-28
refactor: change deploy script output file format

dohaki:
This changes the deploy script to be compatible with the newly created https://github.com/bosonprotocol/instances repository. If we deploy contracts from this repository then the contract addresses now get written to `<ROOT>/addresses/<CHAIN_ID>-<ENV>.json` which are not checked in to git. The outputted file can then be consumed by a script of the https://github.com/bosonprotocol/instances repository to easily update addresses.

This also removes the `contracts-<ENV>.json` files, as they are now maintained in a separate repository.

The change allows the contract deployments to be decoupled from the protocol release and enables a single source of truth for official contract addresses and configurations.


scripts/utils.ts
```diff
@@ -0,0 +1,19 @@
+export const addressesDirPath = __dirname + `/../addresses`;
+
+const availableEnvironments = [
+  'testing',
+  'staging',
+  'production',
+  'hardhat',
+  '',
```
zajck, line 8: Is there any reason, why allow the empty string?

zajck (changes requested):

scripts/utils.ts
dohaki: My initial idea was to keep `--env` optional as it was before. So the default argument of `--env` would be `""` and the output `./addresses/<CHAIN_ID>.json`.

zajck: Makes sense to me.

zajck (approved):

cliffhall (approved):
LGTM! 👍

PR #310, opened 2022-02-10
Add support for ERC721 ownership

hswopeams:
This is a work in progress to add support for ERC721 ownership.

At this moment, the draft PR only includes code changes. The tests have not yet been changed. 
The tests WILL NOT work at this point. This draft PR is to get feedback on the code changes because they are significant.

Also note that during analysis, I suggested using the term "ConditionalCheckType". I decided simply "Condition" was better. "CheckType" is superfluous.

I realized while coding that the `Gate.registerVoucherSetId` and all `BosonRouter.requestCreateOrder*Conditional` functions accept and store a _conditionalTokenId, even though this is not relevant for ERC20s. It's also not relevant for ERC721s if we are only checking the balance. Perhaps I should add some checks in these functions to see if the _conditionalTokenId is necessary/relevant


contracts/Gate.sol
```diff
@@ -151,19 +164,24 @@ contract Gate is IGate, Ownable, Pausable {
...
+     * @param _condition condition that will be checked when a user commits using a conditional token
      */
-    function registerVoucherSetId(uint256 _tokenIdSupply, uint256 _conditionalTokenId)
+    function registerVoucherSetId(uint256 _tokenIdSupply, uint256 _conditionalTokenId, Condition _condition)
         external
         override
         whenNotPaused
         onlyRouterOrOwner
     {
         require(_conditionalTokenId != 0, "TOKEN_ID_0_NOT_ALLOWED");
```
zajck, line 175: I think this should be removed now, since tokenID 0 has a different meaning now.

contracts/Gate.sol
```diff
@@ -217,4 +233,43 @@ contract Gate is IGate, Ownable, Pausable {
...
+     */
+    function checkBalance(address _user, uint256 _tokenIdSupply)
+        internal
+        view
+        returns (bool)
+    {
+        ConditionalCommitInfo memory conditionalCommitInfo = voucherSetToConditionalCommit[_tokenIdSupply];
+        return
+            !isDeactivated[_user][_tokenIdSupply] &&
+            ((conditionalTokenType == TokenType.NONFUNGIBLE_TOKEN || conditionalTokenType == TokenType.FUNGIBLE_TOKEN)
```
zajck, line 250: Instead of checking id it's `NONFUNGIBLE_TOKEN` or `FUNGIBLE_TOKEN` you could just check if it's `MULTI_TOKEN` and switched what should be executed.

contracts/Gate.sol
hswopeams: Removed. Will be included in next commit.

contracts/Gate.sol
hswopeams: Changed. Will be included in next commit.

bidyut-arianelabs (changes requested):

contracts/Gate.sol
```diff
@@ -5,6 +5,7 @@ pragma solidity 0.7.6;
 import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/utils/Pausable.sol";
 import "./interfaces/IGate.sol";
+import ".//UsingHelpers.sol";
```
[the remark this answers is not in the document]
hswopeams: Changed. Will be included in next commit.

hswopeams:
Just closed and reopening now to see if reopening will trigger GitHub Actions. Changing from Draft to open did not

zajck (changes requested):
It looks good to me, just `TOKEN_ID_0_NOT_ALLOWED` can be removed from `revertReasons.ts` since it's nowhere used anymore.
https://github.com/bosonprotocol/contracts/blob/8b45f9a0f91665a29b7cf0bc4aeb0db7f9ee8a42/testHelpers/revertReasons.ts#L67

cliffhall (approved):
LGTM! 👍

bidyut-arianelabs (approved):

zajck (approved):
Since you have two other approvals, I won't block it anymore because of
`TOKEN_ID_0_NOT_ALLOWED` in revertReasons, but please remove that in the next PR.

hswopeams:
> TOKEN_ID_0_NOT_ALLOWE

Yes, will remove in threshold PR.

PR #313, opened 2022-02-16
Add conditional threshold

hswopeams:
This is a draft pull request so code changes can be checked before/while I start on the unit tests.

the Conditional functions in BosonRouter had to be changed to take `ConditionalCommitInfo` struct parameter due to` stack too deep` error that occurred when the `threshold` parameter was first added.

Check the ConditionalCommitInfo to see if it is packed in the most efficient way.

Unit tests will NOT work yet.


contracts/BosonRouter.sol
```diff
@@ -1217,16 +1253,17 @@ contract BosonRouter is
      * @param _condition condition that will be checked when a user commits using a conditional token
      * @param _registerConditionalCommit indicates whether Gate.registerVoucherSetId should be called. Gate.registerVoucherSetId can also be called separately
      */
-    function finalizeConditionalOrder(uint256 _tokenIdSupply, address _gateAddress, uint256 _conditionalTokenId, Condition _condition, bool _registerConditionalCommit) internal {
+    function finalizeConditionalOrder(uint256 _tokenIdSupply, address _gateAddress, uint256 _conditionalTokenId, Condition _condition, uint256 _threshold, bool _registerConditionalCommit) internal {
```
zajck, line 1256: Since `finalizeConditionalOrder` actually takes in all elements of a struct `ConditionalCommitInfo`, it might be sensible that struck as a whole is passed to this function.

`finalizeConditionalOrder(uint256 _tokenIdSupply, ConditionalCommitInfo calldata _conditionalCommitInfo)`

This saves a bit of a code. You then also don't have do destruct `_conditionalCommitInfo` every time you call `finalizeConditionalOrder` from other functions.

hswopeams: Agreed. I thought of this after I pushed the PR. I made this change and will include it in the next commit.

contracts/Gate.sol
```diff
@@ -109,11 +105,12 @@ contract Gate is IGate, Ownable, Pausable {
      * @return conditional token ID if one is associated with a voucher set. Zero could be a valid token ID
      * @return condition that will be checked when a user commits using a conditional token
      */
```
cliffhall, line 107: Need to document that return var

```
@return threshold that may additionally be checked when a user commits using a conditional token
```

cliffhall (changes requested):

contracts/Gate.sol
hswopeams: Added will be included in next commit.

hswopeams:
Closing so can re-open and trigger GitHub Actions

contracts/BosonRouter.sol
```diff
@@ -1247,23 +1232,31 @@ contract BosonRouter is
...
+    function finalizeConditionalOrder(uint256 _tokenIdSupply, ConditionalCommitInfo calldata _conditionalCommitInfo) internal {
+        voucherSetToGateContract[_tokenIdSupply] = _conditionalCommitInfo.gateAddress;
 
-        emit LogConditionalOrderCreated(_tokenIdSupply, _gateAddress, _conditionalTokenId, _condition, _threshold);
+        emit LogConditionalOrderCreated(_tokenIdSupply, _conditionalCommitInfo.gateAddress, _conditionalCommitInfo.conditionalTokenId, _conditionalCommitInfo.condition, _conditionalCommitInfo.threshold);
 
-        if (_registerConditionalCommit) {
-            IGate(_gateAddress).registerVoucherSetId(
+        if (_conditionalCommitInfo.registerConditionalCommit) {
+            IGate(_conditionalCommitInfo.gateAddress).registerVoucherSetId(
```
cliffhall, line 1255: Is there any reason not to just change the `registerVoucherSetId` method to take a `ConditionalCommitInfo` argument rather than passing its first three arguments individually and then creating it with literal defaults for the last two args?

cliffhall (changes requested):

contracts/BosonRouter.sol
hswopeams: No. No real reason. I'll change it.

cliffhall (approved):
LGTM! 👍

zajck (approved):

PR #314, opened 2022-02-18
Dependabot PRs should check deploy script

bidyut-arianelabs:
In this PR, the Deploy script is run in CI to make sure that the new dependabot PRs do not break it. This will make the review process quicker for code reviewers.

The following statement determines if the PR author is Dependabot
```
    if: ${{ github.actor == 'dependabot[bot]' }}
```
Reference here: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/automating-dependabot-with-github-actions#common-dependabot-automations

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -96,5 +96,9 @@ jobs:
           cache: 'npm'
       - name: Install Dependencies
         run: npm install
+      - name: Prepare Environment Variables
+        shell: bash
+        run: |
+          cp .env.example .env
       - name: Deploy contracts locally
         run: npm run contracts:run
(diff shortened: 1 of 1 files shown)


bidyut-arianelabs:
1. We cannot test that this PR will work because dependabot is not the author of this PR.
2. To verify that this PR will NOT run deploy scripts for PRs submitted by Devs, please see https://github.com/bosonprotocol/contracts/actions/runs/1864436101
3. To verify that the deploy scripts integration in CI is done and they do not crash, you can see https://github.com/bosonprotocol/contracts/actions/runs/1864193056

hswopeams (approved):
We won't know for sure until we get another dependabot PR, but looks good from what I can see.

cliffhall (approved):
LGTM! 👍

PR #315, opened 2022-02-22
Track seller per voucher, not pet voucher set

zajck:
This PR resolves the issue of mismatch between voucher seller and escrow, which happened if voucherSet was transfered to the new owner, while some of vouchers were not finalised yet.

Changes:
- added `address seller` to voucherStatus, which is populated when buyer commits to the offer
- added function `getVoucherSeller` which is used instead of `getSupplyHolder` in voucher interactions

### DIFF
--- a/test/2_test_fullpath_with_permit.ts
+++ b/test/2_test_fullpath_with_permit.ts
@@ -4059,7 +4059,7 @@ describe('Cashier and VoucherKernel', () => {
         ).to.be.revertedWith(revertReasons.INEXISTENT_SUPPLY);
       });
 
-      it.only('[!CANCEL] It should not be possible to cancel voucher that does not exist yet', async () => {
+      it('[!CANCEL] It should not be possible to cancel voucher that does not exist yet', async () => {
         // spoof boson router address
         await contractBosonRouter.pause();
         await contractVoucherKernel.setBosonRouterAddress(
--- a/test/2_test_fullpath_with_permit.ts
+++ b/test/2_test_fullpath_with_permit.ts
@@ -5284,7 +5284,7 @@ describe('Cashier and VoucherKernel', () => {
           await utils.redeem(tokenVoucherIdRedeemed, users.buyer.signer);
           await utils.refund(tokenVoucherIdRefunded, users.buyer.signer);
 
-          const transferTx = await utils.safeTransfer1155(
+          await utils.safeTransfer1155(
             users.other1.address,
             users.other2.address,
             tokenSupplyKey,
@@ -5313,6 +5313,7 @@ describe('Cashier and VoucherKernel', () => {
             users.other2.signer
           );
 
+          // new seller should fail
           await expect(
             bosonRouterNewSeller.cancelOrFault(tokenVoucherIdCommited)
           ).to.be.revertedWith(revertReasons.UNAUTHORIZED_COF);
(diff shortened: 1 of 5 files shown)


test/2_test_fullpath_with_permit.ts
```diff
@@ -5252,6 +5252,205 @@ describe('Cashier and VoucherKernel', () => {
...
+          ).to.equal(expectedBalance, 'Wrong withdrawn amount after expiry');
+
+          await utils.withdraw(tokenVoucherIdRedeemed, users.deployer.signer);
+          expectedBalance = expectedBalance.add(
+            constants.PROMISE_DEPOSITSE1 + constants.PROMISE_PRICE1
+          );
+          expect(
+            await contractCashier.getEscrowAmount(users.other1.address)
+          ).to.equal(
+            oldOwnerFromEscrow.sub(2 * constants.PROMISE_DEPOSITSE1),
```
hswopeams, line 5382: Is this correct? Shouldn't it be `oldOwnerFromEscrow.sub(constants.PROMISE_DEPOSITSE1 + constants.PROMISE_PRICE`1) ?

test/2_test_fullpath_with_permit.ts
```diff
@@ -5252,6 +5252,205 @@ describe('Cashier and VoucherKernel', () => {
...
+          expect(
+            await ethers.provider.getBalance(users.other1.address)
+          ).to.equal(expectedBalance, 'Wrong withdrawn amount after redeem');
+
+          await utils.withdraw(tokenVoucherIdRefunded, users.deployer.signer);
+          expectedBalance = expectedBalance.add(constants.PROMISE_DEPOSITSE1);
+          expect(
+            await contractCashier.getEscrowAmount(users.other1.address)
+          ).to.equal(
+            oldOwnerFromEscrow.sub(3 * constants.PROMISE_DEPOSITSE1),
```
hswopeams, line 5394: Why is this `3 * constants.PROMISE_DEPOSITSE1` ?

Is it because `oldOwnerFromEscrow` is defined with `const` and the value never actually gets updated in the previous lines where `oldOwnerFromEscrow.sub` is called?

test/2_test_fullpath_with_permit.ts
```diff
@@ -5252,6 +5252,205 @@ describe('Cashier and VoucherKernel', () => {
...
+          )
+            .to.emit(contractVoucherKernel, eventNames.LOG_CANCEL_VOUCHER_SET)
+            .withArgs(tokenSupplyKey, users.other2.address);
+
+          expect(
+            await contractCashier.getEscrowAmount(users.other2.address)
+          ).to.equal('0');
+        });
+
+        it('After transfer, old seller cannot manage new vouchers anymore', async () => {
```
hswopeams, line 5415: I would remove the "anymore" here since the old seller should never have been able to manage new vouchers from an already transferred voucher set.

test/2_test_fullpath_with_permit.ts
zajck: I think that my answer for comment above explains this as well.

And yes, when calling `oldOwnerFromEscrow.sub` the value of `oldOwnerFromEscrow` actually stays intact/

test/2_test_fullpath_with_permit.ts
zajck: :+1: will change in next commit.

cliffhall (dismissed):
LGTM! 👍

hswopeams (approved):

bidyut-arianelabs (approved):
Looks good to me 👍

PR #317, opened 2022-02-24
Add script to deploy Gate

hswopeams:
This PR adds the deploy-gate.ts script for deploying the Gate contract separately. References to the Gate contract were removed from the deploy.ts and verify.ts scripts.

Note the new .env file properties:
#Gate contract values
BOSON_ROUTER_ADDRESS=0000000000000000000000000000000000000000
CONDITIONAL_TOKEN_ADDRESS=0000000000000000000000000000000000000000
#Must be 0 (FUNGIBLE_TOKEN - ERC20), 1 (NONFUNGIBLE_TOKEN - ERC721), or 2 (MULTI_TOKEN - ERC1155)
CONDITIONAL_TOKEN_TYPE=x

You need to specify an already-deployed Boson Router address (could be one deployed previously or using the deployt.ts script just before running deploy-gate.ts script.

The conditional token address must be the address of an already-deployed token. 

There are some mock token contracts on Rinkeby and Ropsten.

Here are some mock addresses on Rinkeby:
```
{
  "chainId": 4,
  "env": "",
  "erc20": "0x6635ED9a4ab634D1cd98B752039B02dE9796AC40",
  "erc721": "0x9cf9252ba74419c95FA78e45A1f1cA4458B806ab",
  "erc1155": "0x5D9bDC0DA81f661aD32d729dC112C0DEb03B1d48",
  "erc1155NonTransferable": "0x479aCbAE677Dd2831011aCC1D3989CDdF8a36c07"
}
```

See deployer address 0x2a91A0148EE62fA638bE38C7eE05c29a3e568dD8 on Ropsten Etherscan for mock deployments.

Putting the verify action in the same script as the deployment didn't work. The verify function kept failing with a message saying the newly-deployed Gate contract address had not byte code. This is because it takes a little while for the contract to show up in Etherscan. I tried building in some waits, but that didn't help.

I then separated the verify function into a separate script, to be run after the deploy-gate script. I then got an error saying that the contract had already been verified. I ran the deployment script again and checked Etherscan, and indeed it showed the contract as verified, even though I hadn't run the verify script. It's very strange, but it seems to be automatically verifying the Gate contract deployment.

I noticed the same behaviour when deploying the procotol contracts using deploy.ts.

This all applies to Rinkeby. Wasn't able to test Ropsten. Ropsten is being difficult. Transactions aren't getting mined.


zajck (commented):
It looks good to me.
Works on locally forked node.

About the verification, it sort of seems that if the same bytecode is deployed twice, and if it was verified after the first deploymenet, etherscan aucomatically pairs the second deployment with source code.

You can see that VoucherKernel and Cashier are not verified, since the code was changed, and etherscan could not find a match in the existing verified contracts. 
But still this does not explain why BosonRouter (which was also changed) is verified.

I think that we'll still need verfy-gate script to do the verification on other networks.

hswopeams:
> It looks good to me. Works on locally forked node.
> 
> About the verification, it sort of seems that if the same bytecode is deployed twice, and if it was verified after the first deploymenet, etherscan aucomatically pairs the second deployment with source code.
> 
> You can see that VoucherKernel and Cashier are not verified, since the code was changed, and etherscan could not find a match in the existing verified contracts. But still this does not explain why BosonRouter (which was also changed) is verified.
> 
> I think that we'll still need verfy-gate script to do the verification on other networks.

I'll add a verify script

zajck (dismissed):

bidyut-arianelabs (dismissed):

cliffhall (dismissed):
LGTM!👍

zajck (approved):

cliffhall (approved):
LGTM!

PR #321, opened 2022-02-24
Check whether PR author name is dependabot

bidyut-arianelabs:
Earlier I was checking if GHA is run by dependabot. This becomes an issue because the deploy scripts won't run when the GHA jobs are run by any other user. e.g if we need to re-run the GHA jobs.
As a fix, I am checking if PR is submitted by Dependabot now.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -86,7 +86,7 @@ jobs:
   deploy:
     needs: test
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


cliffhall (dismissed):
LGTM!

hswopeams (changes requested):
This didn't solve the problem. We need to close this PR though. We don't have time to fix this now, and it's not a big problem.

hswopeams (approved):

zajck (approved):

PR #334, opened 2022-03-09
VKL-01C: Statement Execution Optimization

### LINKED ISSUE #332: Statement Execution Optimization
## VKL-01C: Statement Execution Optimization

| [Gas Optimization](/appendix/finding-types#gas-optimization) | <span class="o-severity o-informational">Informational</span> | [VoucherKernel.sol:L735](https://github.com/bosonprotocol/contracts/blob/4b966fbe291fe15e1c773fc8dbaa7af2289a7780/contracts/VoucherKernel.sol#L735), [L737-L740](https://github.com/bosonprotocol/contracts/blob/4b966fbe291fe15e1c773fc8dbaa7af2289a7780/contracts/VoucherKernel.sol#L737-L740) |

### Description:

The linked statements will result in no-ops if the `_amount` that is passed to `setDepositsReleased` is zero.

### Example:

```solidity
/**
 * @notice Mark voucher token that the deposits were released
 * @param _tokenIdVoucher   ID of the voucher token
 * @param _to               recipient, one of {ISSUER, HOLDER, POOL}
 * @param _amount           amount that was released in the transaction
 */
function setDepositsReleased(uint256 _tokenIdVoucher, Entity _to, uint256 _amount)
    external
    override
    onlyFromCashier
{
    require(_tokenIdVoucher != 0, "UNSPECIFIED_ID");

    vouchersStatus[_tokenIdVoucher].depositReleased.status |= (ONE << uint8(_to));

    vouchersStatus[_tokenIdVoucher].depositReleased.releasedAmount = vouchersStatus[_tokenIdVoucher].depositReleased.releasedAmount.add(_amount);

    if (vouchersStatus[_tokenIdVoucher].depositReleased.releasedAmount == getTotalDeposits(_tokenIdVoucher)) {
        vouchersStatus[_tokenIdVoucher].isDepositsReleased = true;
        emit LogFundsReleased(_tokenIdVoucher, 1); 
    }        
}
```

### Recommendation:

We advise a conditional to be introduced before the linked statements that will halt execution of the function if the `_amount` is zero.

zajck:
FIx #332 

Three things to note:
- I did not use require, since that would revert transaction. In most cases when `withdraw` is called, at least for some party amount is zero, and imposing that check would render `withdraw` useless
- I considered just putting lines 735 - 740 in if conditional that would execute only if amount > 0 (this is what was proposed). That would result in `vouchersStatus[_tokenIdVoucher].depositReleased.status` still being updated. Although this is not wrong per se, this would result in different states depending on whether `withdraw` is called or if `withdrawSingle` is callled for each entity.
In `withdraw`,  it would seem as if deposit was released to all three entities, while `withdrawSingle` would revert if there was nothing to be withdrawn and therefore `vouchersStatus[_tokenIdVoucher].depositReleased.status` would not be updated. I decided to do it in a way, that calling withdraw or withdrawSIngle three times result in same state.
- I could impose the check `_amount > 0` at very begining of `setDepositsReleased` and just return if `amount ==0`, but I decided to actually put that check into `Cashier` to avoid a call to VoucherKernel contract at all. I added a comment in `setDepositsReleased` so if at any point  `onlyFromCashier` is removed that amount tcheck should be put in place.

### DIFF
--- a/contracts/Cashier.sol
+++ b/contracts/Cashier.sol
@@ -402,10 +402,11 @@ contract Cashier is ICashier, ReentrancyGuard, Ownable, Pausable {
 
         _released = _amount > 0;
 
-        IVoucherKernel(voucherKernel).setDepositsReleased(
-            _voucherDetails.tokenIdVoucher, _to, _amount
-        );
-
+        if (_released) {
+            IVoucherKernel(voucherKernel).setDepositsReleased(
+                _voucherDetails.tokenIdVoucher, _to, _amount
+            );
+        }
     }
 
     /**
--- a/contracts/VoucherKernel.sol
+++ b/contracts/VoucherKernel.sol
@@ -719,6 +719,7 @@ contract VoucherKernel is IVoucherKernel, Ownable, Pausable, ReentrancyGuard {
 
     /**
      * @notice Mark voucher token that the deposits were released
+     * @dev    Currently Cashier makes a check that _amount > 0. If onlyFromCashier is ever removed, this function should check that _amount > 0
      * @param _tokenIdVoucher   ID of the voucher token
      * @param _to               recipient, one of {ISSUER, HOLDER, POOL}
      * @param _amount           amount that was released in the transaction


bidyut-arianelabs (approved):
Looks good to me 👍

hswopeams (approved):

PR #335, opened 2022-03-09
Remediate GAT-02C: Sub-Optimal Storage Read

### LINKED ISSUE #336: Sub-Optimal Storage Read
## GAT-02C: Sub-Optimal Storage Read

| [Gas Optimization](/appendix/finding-types#gas-optimization) | <span class="o-severity o-informational">Informational</span> | [Gate.sol:L111](https://github.com/bosonprotocol/contracts/blob/4b966fbe291fe15e1c773fc8dbaa7af2289a7780/contracts/Gate.sol#L111) |

### Description:

The linked statement reads all members of the `ConditionalCommitInfo` struct from storage yet only three are utilized.

### Example:

```solidity
/**
 * @notice Get the token ID and Condition associated with the supply token ID (voucherSetID)
 * @param _tokenIdSupply an ID of a supply token (ERC-1155) [voucherSetID]
 * @return conditional token ID if one is associated with a voucher set. Zero could be a valid token ID
 * @return condition that will be checked when a user commits using a conditional token
 * @return threshold that may be checked when a user commits using a conditional token
 */
function getConditionalCommitInfo(uint256 _tokenIdSupply) external view returns (uint256, Condition, uint256) {
    ConditionalCommitInfo memory conditionalCommitInfo = voucherSetToConditionalCommit[_tokenIdSupply];
    return (
        conditionalCommitInfo.conditionalTokenId,
        conditionalCommitInfo.condition,
        conditionalCommitInfo.threshold
    );
}
```

### Recommendation:

We advise the declaration to be set as `storage` optimizing the reads the function performs.

hswopeams:
This is a view function that is NOT called internally, so there is no gas cost. I'm making this change just to satisfy the auditors.

### DIFF
--- a/contracts/Gate.sol
+++ b/contracts/Gate.sol
@@ -108,7 +108,7 @@ contract Gate is IGate, Ownable, Pausable {
      * @return threshold that may be checked when a user commits using a conditional token
      */
     function getConditionalCommitInfo(uint256 _tokenIdSupply) external view returns (uint256, Condition, uint256) {
-        ConditionalCommitInfo memory conditionalCommitInfo = voucherSetToConditionalCommit[_tokenIdSupply];
+        ConditionalCommitInfo storage conditionalCommitInfo = voucherSetToConditionalCommit[_tokenIdSupply];
         return (
             conditionalCommitInfo.conditionalTokenId,
             conditionalCommitInfo.condition,


bidyut-arianelabs (approved):
Looks good to me 👍

zajck (approved):
LGTM

PR #338, opened 2022-03-09
UHS-01C: Potential Tight-Packing Optimization

### LINKED ISSUE #331: Potential Tight-Packing Optimization
## UHS-01C: Potential Tight-Packing Optimization

| [Gas Optimization](/appendix/finding-types#gas-optimization) | <span class="o-severity o-informational">Informational</span> | [UsingHelpers.sol:L55-L61](https://github.com/bosonprotocol/contracts/blob/4b966fbe291fe15e1c773fc8dbaa7af2289a7780/contracts/UsingHelpers.sol#L55-L61) |

### Description:

THe `DepositsReleased` struct contains two variable declarations in sequence one of which is a `uint8` and the other being a `uint256`.

### Example:

```solidity
struct DepositsReleased {
    // bool issuer;
    // bool holder;
    // bool pool;
    uint8 status;
    uint256 releasedAmount;
}
```

### Recommendation:

We advise the capability of the `uint256` to be downgraded to be assessed and if possible to downgrade it to a `uint248` variable permitting both values to be packed to a single storage slot and thus cause a single `SSTORE` operation to be executed when both values are written to.

zajck:
Close #331 

Beside the struct change, I needed to adjust calculation at one point.

Could not use `SafeMath` for `uint248` directly, since it returns `uint256`, so two explicit conversions are added

### DIFF
--- a/contracts/UsingHelpers.sol
+++ b/contracts/UsingHelpers.sol
@@ -53,11 +53,8 @@ struct VoucherStatus {
 }
 
 struct DepositsReleased {
-    // bool issuer;
-    // bool holder;
-    // bool pool;
     uint8 status;
-    uint256 releasedAmount;
+    uint248 releasedAmount;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


hswopeams (approved):

bidyut-arianelabs (approved):
Looks good to me 👍

PR #341, opened 2022-03-10
Remediate VKL-01S

### LINKED ISSUE #325: Redundant Local Variable Declaration
## VKL-01S: Redundant Local Variable Declaration

| [Gas Optimization](/appendix/finding-types#gas-optimization) | <span class="o-severity o-informational">Informational</span> | [VoucherKernel.sol:L206](https://github.com/bosonprotocol/contracts/blob/4b966fbe291fe15e1c773fc8dbaa7af2289a7780/contracts/VoucherKernel.sol#L206), [L207](https://github.com/bosonprotocol/contracts/blob/4b966fbe291fe15e1c773fc8dbaa7af2289a7780/contracts/VoucherKernel.sol#L207) |

### Description:

The linked local variable declarations can be substituted with a contract-level `constant`.

### Example:

```solidity
uint256 _complainPeriod = 7 * 1 days;
uint256 _cancelFaultPeriod = 7 * 1 days;
complainPeriod = _complainPeriod;
cancelFaultPeriod = _cancelFaultPeriod;

emit LogComplainPeriodChanged(_complainPeriod, msg.sender);
emit LogCancelFaultPeriodChanged(_cancelFaultPeriod, msg.sender);
```

### Recommendation:

We advise one to be declared to optimize the gas cost of the contract's construction.

hswopeams:
The remediation is different from what is suggested in the audit, but it makes the code more legible. The complainPeriod and cancelFaultPeriod can't be constants (as suggested by the auditors) because they need to be settable using the `setComplainPeriod` and `setCancelFaultPeriod` functions. These functions are needed so that the periods can be set to lower values for testing on testnet. Otherwise, we would have to wait two weeks to test some scenarios.  

As discussed, the set functions are now being called from the constructor. The set functions emit the events. Normally when calling a function from within the same contract, the function visibility is public. It is also necessary to change the set function visibility to public because it's not possible to call an external function on the same contract from the constructor.I made the function visibility change, but this necessitated removing the functions from the IVoucherKernel interface. I ran the gas reporter in the before and after situation by executing two test cases -- one that tests `setComplainPeriod` and one that calls `setCancelPeriod`. Both test cases cause the constructor to be called because the contracts are redeployed before each test case.

The result is that after this change, the gas used by the constructor goes UP by 4868. I've attached the gas reports. The GasReport-external-setters-not-called-from-constructor.txt represents the original situation before the change. Based on the gas reports and the fact that the issue pointed out by the auditors is to optimize gas, I'm not sure it is worth making this change. On the other hand, the constructor code is simplified.

[GasReport-external-setters-not-called-from-constructor.txt](https://github.com/bosonprotocol/contracts/files/8221611/GasReport-external-setters-not-called-from-constructor.txt)
[GasReport-public-setters-called-from-constructor.txt](https://github.com/bosonprotocol/contracts/files/8221612/GasReport-public-setters-called-from-constructor.txt)

### DIFF
--- a/contracts/interfaces/IVoucherKernel.sol
+++ b/contracts/interfaces/IVoucherKernel.sol
@@ -152,6 +152,12 @@ interface IVoucherKernel {
     ) external;
 
     /**
+     * @notice Set the general cancelOrFault period, should be used sparingly as it has significant consequences. Here done simply for demo purposes.
+     * @param _cancelFaultPeriod   the new value for cancelOrFault period (in number of seconds)
+     */
+    function setCancelFaultPeriod(uint256 _cancelFaultPeriod) external;
+
+    /**
      * @notice Set the address of the Boson Router contract
      * @param _bosonRouterAddress   The address of the BR contract
      */
@@ -175,6 +181,12 @@ interface IVoucherKernel {
      */
     function setVoucherSetTokenAddress(address _voucherSetTokenAddress)
         external;
+
+    /**
+     * @notice Set the general complain period, should be used sparingly as it has significant consequences. Here done simply for demo purposes.
+     * @param _complainPeriod   the new value for complain period (in number of seconds)
+     */
+    function setComplainPeriod(uint256 _complainPeriod) external;
 
     /**
      * @notice Get the promise ID at specific index
--- a/contracts/interfaces/IVoucherKernel.sol
+++ b/contracts/interfaces/IVoucherKernel.sol
@@ -152,12 +152,6 @@ interface IVoucherKernel {
     ) external;
 
     /**
-     * @notice Set the general cancelOrFault period, should be used sparingly as it has significant consequences. Here done simply for demo purposes.
-     * @param _cancelFaultPeriod   the new value for cancelOrFault period (in number of seconds)
-     */
-    function setCancelFaultPeriod(uint256 _cancelFaultPeriod) external;
-
-    /**
      * @notice Set the address of the Boson Router contract
      * @param _bosonRouterAddress   The address of the BR contract
      */
@@ -181,12 +175,6 @@ interface IVoucherKernel {
      */
     function setVoucherSetTokenAddress(address _voucherSetTokenAddress)
         external;
-
-    /**
-     * @notice Set the general complain period, should be used sparingly as it has significant consequences. Here done simply for demo purposes.
-     * @param _complainPeriod   the new value for complain period (in number of seconds)
-     */
-    function setComplainPeriod(uint256 _complainPeriod) external;
 
     /**
      * @notice Get the promise ID at specific index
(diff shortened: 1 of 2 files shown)


zajck:
What if don't actually declare local variables and just set:
```
setComplainPeriod(7 * 1 days);
setCancelFaultPeriod(7 * 1 days);
```
That would save some gas. But yeah, overall it will still be more expensive.

Or alternatively make a contract level  constant
```
uint256 internal constant WEEK = 7 * 1 days;
```
and then
```
setComplainPeriod(WEEK);
setCancelFaultPeriod(WEEK);
```
This should have comparable cost to my other suggestion.

In any case, I think that simplicity is better than gas cost optimization since it's one time cost. I'm supporting the change.

hswopeams:
I added the WEEK constant, as suggested by @zajck.
It saves a very small amount of gas (16) but also makes the code cleaner. 
[GasReport-public-setters-called-from-constructor-with-constant.txt](https://github.com/bosonprotocol/contracts/files/8222033/GasReport-public-setters-called-from-constructor-with-constant.txt)

contracts/interfaces/IVoucherKernel.sol
```diff
@@ -151,12 +151,6 @@ interface IVoucherKernel {
         address _newSeller
     ) external;
 
-    /**
-     * @notice Set the general cancelOrFault period, should be used sparingly as it has significant consequences. Here done simply for demo purposes.
-     * @param _cancelFaultPeriod   the new value for cancelOrFault period (in number of seconds)
-     */
-    function setCancelFaultPeriod(uint256 _cancelFaultPeriod) external;
```
zajck, line 158: You should leave this in the interface. No problem if it's external here, but the actual implementation is public.

contracts/interfaces/IVoucherKernel.sol
```diff
@@ -182,12 +176,6 @@ interface IVoucherKernel {
     function setVoucherSetTokenAddress(address _voucherSetTokenAddress)
         external;
 
-    /**
-     * @notice Set the general complain period, should be used sparingly as it has significant consequences. Here done simply for demo purposes.
-     * @param _complainPeriod   the new value for complain period (in number of seconds)
-     */
-    function setComplainPeriod(uint256 _complainPeriod) external;
-
```
zajck, line 190: You should leave this in the interface. No problem if it's external here, but the actual implementation is public.

zajck (changes requested):
That's cool.

I forgot to submit before my comment about the interface, so I'm doing it now.

contracts/interfaces/IVoucherKernel.sol
hswopeams: Added back in.

contracts/interfaces/IVoucherKernel.sol
hswopeams: Added back in.

zajck (approved):

bidyut-arianelabs (approved):
Looks good to me 👍

PR #342, opened 2022-03-10
VoucherKernel.getTotalDeposits fix

zajck:
`VoucherKernel.getTotalDeposits` expected that `tokenIdSupply` is passed in, but there was a mistake and the `voucherId` was passed in.

I renamed the function to `getTotalDepositsForVoucher` to be clear it depends on voucher, not on voucher set and modified it to return correct values.

A consequence of bug was, that `LogFundsReleased` was not emitted when it should be. I noticed that this was not properly tested so I also added missing checks to `3_withdrawals` and `3_1_withdrawals_per_entity`

### DIFF
--- a/test/3_1_withdrawals_per_entity.ts
+++ b/test/3_1_withdrawals_per_entity.ts
@@ -478,10 +478,7 @@ describe('Cashier withdrawals ', () => {
 
         const recipients = [...depositRecipients, ...paymentRecipients];
 
-        // let lastDepositRecepient = isEntityRecipient(
-        //   [...depositRecipients, ...paymentRecipients],
-        //   Entity[-1]
-        // )
+        // determine who is the last entity that should receive at leas some part of the deposit
         let lastDepositRecepient;
         for (const entity of [...Object.keys(Entity)].reverse()) {
           if (isEntityRecipient(recipients, entity)) {
--- a/test/3_1_withdrawals_per_entity.ts
+++ b/test/3_1_withdrawals_per_entity.ts
@@ -399,6 +399,16 @@ describe('Cashier withdrawals ', () => {
                   }
                 );
 
+                eventUtils.assertEventEmitted(
+                  txReceipt,
+                  VoucherKernel_Factory,
+                  eventNames.LOG_FUNDS_RELEASED,
+                  (ev) => {
+                    assert.isTrue(ev._tokenIdVoucher.eq(voucherID));
+                    assert.equal(ev._type, 0); // 0 == withdraw
+                  }
+                );
+
                 if (ethTransfers && paymentWithdrawal) {
                   // only eth transfers emit LOG_WITHDRAWAL. If price was in TKN, paymentWithdrawal == null and no adjustment is needed
                   eventUtils.assertEventEmitted(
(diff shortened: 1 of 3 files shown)


test/3_withdrawals.ts
```diff
@@ -454,6 +464,31 @@ describe('Cashier withdrawals ', () => {
           }
         );
 
+        if (i == 0) {
```
krzysztofpaliga, line 467: sorry, I have a difficulty to follow. What does this "i" stand for?

test/3_withdrawals.ts
```diff
@@ -454,6 +464,31 @@ describe('Cashier withdrawals ', () => {
...
+              assert.isTrue(ev._tokenIdVoucher.eq(voucherID));
+
+              assert.equal(ev._type, expectedType++); // 0 == withdraw, 1 == deposit
+            }
+          );
+        } else {
+          eventUtils.assertEventEmitted(
+            txReceipt,
+            VoucherKernel_Factory,
+            eventNames.LOG_FUNDS_RELEASED,
```
krzysztofpaliga, line 484: I might miss the context, but if Event Type is 1 (deposit), should we use eventNames.LOG_FUNDS_RELEASED (sounds to me like a withdrawal)?

test/3_withdrawals.ts
zajck: This will probably be easier if we get on a call to explain what is happening in this file. But to put it shortly:

Voucher goes through different states - COMMIT, REDEEM, COMPAIN etc...
You can call technically call withdraw no matter in which state it is, but the outcome is not always the same.
There are three possible payouts:
- if withdraw is called before redeem/refund/exipre nothing is withdraw
- if it is called bewteen redeem/refund/exipre and finalize, only payment is withdrawn
- if it is called after finalize there are two options
  - if withdraw was already called before, only deposits are withdrawn
  - if withdraw was not called before, both deposits and payment are withdrawn
  
Now to test all this possible scenarios, I came up with that procedure, where you put in the states through voucher will go and it then just checks that withdraw behaves as expected in each possible scenario that could happen.

So `i` here represents the index of the possible scenarion. As per design `i==0` is always the scenario, where everybody waits until very end to call the withdraw, resulting in both payments and deposits being withdraw at same time. So in this case events emitted and state changes are a bit different then in any other scenarios and are thus in this conditional.

test/3_withdrawals.ts
zajck: `LOG_FUNDS_RELEASED` is ok, since we are actually releasing funds and withdrawing them.

Whenever we are releasing them, those funds can have one of two possible origins:
- funds were sent to the escrow as a payment for the product
- funds were sent to the escrow as a buyer or seller deposit, which are there to prevent misbehaviour

So basicaly `eventType` here just tells what the origin of withdrawn funds is.
But yes, I made a mistake in the comments, since 0 does not represent withdraw but "payment". I will correct that.

test/3_1_withdrawals_per_entity.ts
```diff
@@ -466,13 +476,22 @@ describe('Cashier withdrawals ', () => {
           );
         }
 
+        const recipients = [...depositRecipients, ...paymentRecipients];
+
+        // let lastDepositRecepient = isEntityRecipient(
+        //   [...depositRecipients, ...paymentRecipients],
+        //   Entity[-1]
+        // )
```
hswopeams, line 484: Are these comments meant to be here?

zajck: Nope. Removed.

hswopeams (approved):

ghost (approved):

bidyut-arianelabs (approved):
Looks good to me 👍

PR #343, opened 2022-03-10
Deployment script without ERC1155NonTransferable

zajck:
- old script was renamed to `deploy-legacy`
- new `deploy` is more or less the same, expect all recerences to ERC1155 are removed
- I added another script to deploy ERC1155NonTransferable. Compared to `deploy-mocks` this script read metadata uri from .env and properly put it in the constructor
- I added the corresponding tasks, but I did not make additional scripts in `package.json` since I think it will become too clutered. Users can still call them using `npx hardhat [taskname]`

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,21 +110,21 @@ This makes the BOSON test token deployed on Rinkeby (0xEDa08eF1c6ff51Ca7Fd681295
 
 #### Special deployments
 
-The following instructions assume you have started a hardhat node (forked or not) using the local node instructions above.  
+The following instructions assume you have started a hardhat node (forked or not) using the local node instructions above. If you want to deploy the contracts to testnets or mainnet just replace `localhost` with the desired network. 
 Calling `npx hardhat deploy` will deploy the protocol contracts. Beside that we provide addtional utility script for the following deployment cases:
 
 - Deploy all protocol contracts and ERC1155NonTransferable (equivalent to v1.0 deployment script)
 
   ```
-  npx hardhat deploy-with-erc1155
+  npx hardhat deploy-with-erc1155 --network localhost
   ```
 - Deploy only ERC1155NonTransferable and set the metadata uri defined in `.env` file
   ```
-  npx hardhat deploy-erc1155-only
+  npx hardhat deploy-erc1155-only --network localhost
   ```
 - Deploy a set of mock token contracts ERC20, ERC721, ERC1155 and ERC1155NonTransferable which can be used to test conditional commit
   ```
-  npx hardhat deploy-mocks
+  npx hardhat deploy-mocks --network localhost
   ```
 - Deploy a Gate contract on a provided network
   - First run the deploy task to deploy the Boson Protocol contracts to your local node, as described above
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ This makes the BOSON test token deployed on Rinkeby (0xEDa08eF1c6ff51Ca7Fd681295
 
 #### Special deployments
 
+The following instructions assume you have started a hardhat node (forked or not) using the local node instructions above.  
 Calling `npx hardhat deploy` will deploy the protocol contracts. Beside that we provide addtional utility script for the following deployment cases:
 
 - Deploy all protocol contracts and ERC1155NonTransferable (equivalent to v1.0 deployment script)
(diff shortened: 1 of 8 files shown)


scripts/deploy-erc1155-non-transferable.ts
```diff
@@ -0,0 +1,117 @@
+//AssetRegistry not used in demo-app
+//const AssetRegistry = artifacts.require("AssetRegistry");
```
bidyut-arianelabs, line 2: Can we please remove these lines from these files?
1. scripts/deploy-erc1155-non-transferable.ts
2. scripts/deploy-protocol-with-ERC1155.ts
3. scripts/deploy-gate.ts

bidyut-arianelabs (changes requested):

scripts/deploy-erc1155-non-transferable.ts
zajck: Done.
I also renamed of the the files but forgot to check that in before, so pull again if you want to test scripts locally.

bidyut-arianelabs: Can we please remove from `scripts/deploy-gate.ts` as well?

zajck: sorry, missed that one. I checked it in now.

bidyut-arianelabs (dismissed):
Looks good to me 👍

README.md
```diff
@@ -103,6 +107,29 @@ npx hardhat deploy --network localhost
...
+  ```
+- Deploy only ERC1155NonTransferable and set the metadata uri defined in `.env` file
+  ```
+  npx hardhat deploy-erc1155-only
+  ```
+- Deploy a set of mock token contracts ERC20, ERC721, ERC1155 and ERC1155NonTransferable which can be used to test conditional commit
+  ```
+  npx hardhat deploy-mocks
+  ```
+- Deploy a Gate contract on a provided network
```
hswopeams, line 128: The gate deployment script won't work properly if the .env file isn't filled out properly. I suggest to change this section to the following markup:

- Deploy a Gate contract on a provided network
* First run either the deploy-with-erc1155 or deploy task to deploy the Boson Protocol contracts locally, as described above
* Then run deploy-mocks, as described above
* Provide values for the folowing properties in your .env file
  ```
  BOSON_ROUTER_ADDRESS=0000000000000000000000000000000000000000
  CONDITIONAL_TOKEN_ADDRESS=0000000000000000000000000000000000000000
  #Must be 0 (FUNGIBLE_TOKEN - ERC20), 1 (NONFUNGIBLE_TOKEN - ERC721), or 2 (MULTI_TOKEN - ERC1155)
  CONDITIONAL_TOKEN_TYPE=x

  ```
  The Boson Router address and conditional token address can be found in the corresponding file in the `addresses` directory. For local deployments, Boson Router will be in a filed called `1.json` (deploy), or `1-legacy.json` (deploy-with-erc1155). The mock conditional token addresses can be found in `1-mocks.json`

  Then run
  ```
  npx hardhat deploy-gate
  ```
Some of the markup formatting might have been lost when pasting it here.

scripts/deploy-erc1155-non-transferable.ts
```diff
@@ -0,0 +1,113 @@
...
+    this.txOptions = {
+      maxPriorityFeePerGas: this.maxTip,
+      gasLimit: 4500000, // our current max contract is around 4.2m gas
+    };
+  }
+
+  async deployContracts() {
+    const signers = await ethers.getSigners();
+    let [, ccTokenDeployer] = signers;
+    const [primaryDeployer] = signers;
```
hswopeams, line 37: I don't think this script needs the primaryDeployer variable.

hswopeams (changes requested):

scripts/deploy-erc1155-non-transferable.ts
zajck: It was easier (=lazier) to keep it in, since it handeled properly both cases when cc and primary are the same or different.

But I updated it now to properly reflect what's going on.

hswopeams (dismissed):

bidyut-arianelabs (approved):

hswopeams (approved):

PR #349, opened 2022-03-14
ERC1155NonTransferable deploy fixes

zajck:
Three ERC1155NonTransferable deploy changes were implemented:
- Script `deploy-erc1155-non-transferable` did not properly store `env` value, which resulted in wrong output file name and wrong content. This is fixed now.
- Removed verification of `ERC1155NonTransferable` in `verify.ts` and made a separate procedure in `verify-erc1155-non-transferable`. Also added hardhat task for that.
- Added `ERC1155NONTRANSFERABLE_TOKEN` to `.env` fail, so if the contract already exist at the time of deployment of protocol contracts, its address can be inclded in `addresses` file


hswopeams (approved):
Can you add some scripts to package.json for deploying and verifying only ERC1155NonTransferable?

hswopeams (changes requested):
Can you add some scripts to package.json for deploying and verifying only ERC1155NonTransferable?

hswopeams (approved):

bidyut-arianelabs (approved):

PR #350, opened 2022-03-14
Modify verify-gate.ts to handle multiple gates

hswopeams:
This PR reuses some types from the instances repo


zajck (approved):
Looks good to me.

The only non-optimal thing is that it always tries to verfiy everything, so if everything is verified already and I add a new gate, this will re-verify the old ones and get Already verified error. But since it is properly handeled and it does not break the script, I approve.

bidyut-arianelabs (approved):

PR #351, opened 2022-03-14
Verify that deployed address matches calculated address

zajck:
In `deploy` we calculate some address upfront and pass them into constructors of other contracts.
If there exist some pending transactions from the deployer at the the begining of the deployment, calculated addresses are wrong.

Previously we only checked at the end of the deployment if calculated addresses actually match the deployed addresses.

Now I changed it to check right after the deployment and terminate the process if mismatch is detected.

### DIFF
--- a/scripts/deploy-legacy.ts
+++ b/scripts/deploy-legacy.ts
@@ -30,6 +30,7 @@ class DeploymentExecutor {
   cancelFaultPeriod;
   maxTip;
   txOptions;
+  contractAddresses;
 
   constructor() {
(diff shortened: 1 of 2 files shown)


hswopeams (changes requested):
I think this is a really good addition to the script. Please add it to` deploy-legacy.ts` as well

zajck:
> I think this is a really good addition to the script. Please add it to` deploy-legacy.ts` as well

Good point. Added.

cliffhall (approved):
LGTM! 👍

hswopeams (approved):

